PR #1, opened 2021-01-20
Fix batch variable generation

t-imamichi:
### Summary

Fixes https://github.com/Qiskit/qiskit-aqua/issues/1462
Copy of https://github.com/Qiskit/qiskit-aqua/pull/1507

This PR moves `keys` argument first to match with [docplex interface](http://ibmdecisionoptimization.github.io/docplex-doc/mp/docplex.mp.model.html?highlight=solve_details#docplex.mp.model.Model.binary_var_dict). `keys` should be given explicitly because the methods are to generate multiple variables at once. `keys=None` does not make sense.

These methods are introduced only in master of Aqua and not yet included in the released version.

### Details and comments

- [X] Simplify the internal of variable generation.
- [X] Avoid an infinite loop due to variable index search.
- [x] Cleanup of unit tests.
- [x] Lint

### DIFF
--- a/qiskit_optimization/problems/quadratic_program.py
+++ b/qiskit_optimization/problems/quadratic_program.py
@@ -12,30 +12,33 @@
 
 """Quadratic Program."""
 
-from typing import cast, List, Union, Dict, Optional, Tuple
 import logging
+import warnings
 from collections import defaultdict
 from collections.abc import Sequence
 from enum import Enum
 from math import fsum, isclose
-import warnings
-import numpy as np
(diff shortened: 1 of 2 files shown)


t-imamichi:
@adekusar-drl I copied https://github.com/Qiskit/qiskit-aqua/pull/1507 and fixed a corner case that causes an infinite loop. Could you take a look at it?

adekusar-drl (approved):

stefan-woerner (approved):

manoelmarques (approved):

PR #6, opened 2021-01-25
Add cplex for python 3.8

t-imamichi:
### Summary

Add cplex 20.1 for Python 3.8.
https://pypi.org/project/cplex/20.1.0.0/#files

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ setuptools.setup(
     include_package_data=True,
     python_requires=">=3.6",
     extras_require={
-        'cplex': ["cplex; python_version >= '3.6' and python_version < '3.8'"],
+        'cplex': ["cplex; python_version >= '3.6' and python_version < '3.9'"],
     },
     zip_safe=False
 )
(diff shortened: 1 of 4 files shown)


manoelmarques (approved):

t-imamichi:
Thank you @manoelmarques for updating github actions. The unit tests with cplex and python 3.8 pass.

setup.py
```diff
@@ -69,7 +69,7 @@
     include_package_data=True,
     python_requires=">=3.6",
     extras_require={
-        'cplex': ["cplex; python_version >= '3.6' and python_version < '3.8'"],
+        'cplex': ["cplex; python_version >= '3.6' and python_version < '3.9'"],
```
woodsp-ibm, line 72: A similar change was done when supporting 3.9 in Aqua. I will note that since we only support Python 3.6 or above, as per the python requires, the python_version >= 3.6 is really not needed here (in Aqua it just has < 3.9 for this entry

PR #17, opened 2021-02-01
Remove DeprecationWarning of np.bool

manoelmarques:


adekusar-drl:
please, please, merge it, the code litters a lot now.

test/test_qaoa.py
```diff
@@ -318,7 +318,7 @@ def test_qaoa_random_initial_point(self):
         qaoa = QAOA(optimizer=NELDER_MEAD(disp=True), p=1, quantum_instance=q_i)
         _ = qaoa.compute_minimum_eigenvalue(operator=qubit_op)
 
-        np.testing.assert_almost_equal([-0.8792, 0.3948], qaoa.optimal_params, decimal=4)
+        np.testing.assert_almost_equal([-0.2372, 0.3423], qaoa.optimal_params, decimal=3)
```
woodsp-ibm, line 321: Is this because random changed with new version? I see this equivalent was altered in Terra. Which has me asking the question why do we have a unit test for QAOA at this level. I can see using it as part of some other testing, but a unit test targeted at QAOA I think we should just the one in Terra no?

manoelmarques: It because of changes in QAOA in Terra. I can eliminate this file altogether since there is one in Terra now

adekusar-drl: Ah, the same test here causes the same problem... Could it be even better if we delete it? Seems like it is not very stable.

manoelmarques: ok, I will remove it

woodsp-ibm (approved):

Issue #8, opened 2020-11-16
Warm Start QAOA


adekusar-drl:
### Warm Start QAOA
Add support for warm start QAOA as in the paper: https://arxiv.org/abs/2009.10095

PR #18, opened 2021-02-01
Implement warm start QAOA optimizer

adekusar-drl:
### Summary
Implement as per #8.

### Details and comments
Should be merged after https://github.com/Qiskit/qiskit-terra/pull/5678


qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
```diff
@@ -0,0 +1,339 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# todo: update documentation
```
woodsp-ibm, line 13: Todo...

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
```diff
@@ -0,0 +1,339 @@
...
+from qiskit_optimization.algorithms import MinimumEigenOptimizationResult, MinimumEigenOptimizer, \
+    OptimizationAlgorithm, OptimizationResultStatus, SolutionSample
+from qiskit_optimization.converters import QuadraticProgramConverter
+from qiskit_optimization.problems import VarType
+
+
+class BaseAggregator(ABC):
+    """A base abstract class for aggregates results"""
+
+    def aggregate(self, results: List[MinimumEigenOptimizationResult]) \
```
woodsp-ibm, line 35: Ought to have `@abtractmethod` decorator in this class I think

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
```diff
@@ -0,0 +1,339 @@
...
+        for state in dict_samples:
+            # sample = (state, dict_samples[state][0], dict_samples[state][1] / num_results)
+            sample = SolutionSample(x=_from_string(state),
+                                    fval=dict_samples[state][0],
+                                    probability=dict_samples[state][1] / num_results,
+                                    status=OptimizationResultStatus.SUCCESS)
+            aggregated_samples.append(sample)
+
+        # new_samples.sort(key=lambda sample: problem.objective.sense.value * sample[1])
+        # x = [float(e) for e in new_samples[0][0]]
```
woodsp-ibm, line 94: Should we just get rid of this code if its not needed

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
```diff
@@ -0,0 +1,339 @@
...
+
+            circuit.ry(-theta, index)
+            circuit.rz(-2.0 * beta, index)
+            circuit.ry(theta, index)
+
+        return circuit
+
+
+class WarmStartQAOAOptimizer(MinimumEigenOptimizer):
+    # todo: update documentation
```
woodsp-ibm, line 184: Todo...

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
```diff
@@ -0,0 +1,339 @@
...
+                                    status=OptimizationResultStatus.SUCCESS)
+            aggregated_samples.append(sample)
+
+        # new_samples.sort(key=lambda sample: problem.objective.sense.value * sample[1])
+        # x = [float(e) for e in new_samples[0][0]]
+
+        return aggregated_samples
+
+
+class WarmStartQAOACircuitFactory:
```
woodsp-ibm, line 99: Hopefully people don't get the wrong idea with the name CircuitFactory in here. We have had CircuitFactory in Aqua for some time that was only finally deprecated for removal at the last release. As this Factory also generates a list of floats would it be better called simply WarmStartQAOAFactory or WarmStartQAOAVariablesFactory.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
```diff
@@ -0,0 +1,339 @@
...
+        """
+        Args:
+            epsilon: the regularization parameter that changes the initial variables according to
+                xi = epsilon if xi < epsilon
+                xi = 1-epsilon if xi > epsilon.
+                The regularization parameter epsilon should be between 0 and 0.5. When it
+                is 0.5 then warm start corresponds to standard QAOA.
+        Raises:
+            AquaError: if ``epsilon`` is not specified for the warm start QAOA or value is not in
+                the range [0, 0.5].
```
woodsp-ibm, line 116: I do not see it raising AquaError - but just noting that in Optimization repo we should be raising an QiskitOptimizationError now, but of course we may pass on exceptions from Terra unaltered but there Aqua Error is removed too

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
```diff
@@ -0,0 +1,339 @@
...
+            (either an explicit value or the default one), then an instance of
+            :class:`~qiskit.optimization.algorithms.WarmStartQAOACircuitFactory` is created.
+            If ``circuit_factory`` is specified then this instance is used in the implementation
+            and ``epsilon`` value is ignored.
+
+        Args:
+            pre_solver: An instance of an optimizer to solve the relaxed version of the problem.
+            relax_for_pre_solver: True if the problem must be relaxed to the continuous case
+                before passing it to the pre-solver.
+            qaoa: A QAOA instance to be used in the computations.
```
woodsp-ibm, line 211: Very specifically we need an instance of the QAOA algorithm from Terra right since it will be updated and we expect it to have certain setters - at least if anyone created their own QAOA it would have to conform in that regard

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
```diff
@@ -0,0 +1,339 @@
...
+            converters: The converters to use for converting a problem into a different form.
+                By default, when None is specified, an internally created instance of
+                :class:`~qiskit.optimization.converters.QuadraticProgramToQubo` will be used.
+
+        Raises:
+            QiskitOptimizationError: if ``epsilon`` is not in the range [0, 0.5].
+        """
+        if epsilon < 0. or epsilon > 0.5:
+            raise QiskitOptimizationError(
+                'Epsilon for warm-start QAOA needs to be between 0 and 0.5.')
```
woodsp-ibm, line 236: It would be nice to include the actual epsilon value that was received in the error message

setup.py
```diff
@@ -70,6 +70,7 @@
     python_requires=">=3.6",
     extras_require={
         'cplex': ["cplex; python_version < '3.9'"],
+        'cvx': ['cvxpy>1.0.0,!=1.1.0,!=1.1.1,!=1.1.2'],
```
woodsp-ibm, line 73: Is this cvx version list what is needed for the logic here or was it just inherited from what we had before that I think was done to be the same as elsewhere in qiskit. I think I would set this to what will work for the code here and let the install process figure an apprropriate version given other cvx dependencies. I just looked at Ignis and its has a version dependency but it simply needs one greater than some specific version.

qiskit_optimization/algorithms/goemans_williamson_optimizer.py
```diff
@@ -0,0 +1,254 @@
...
+from qiskit_optimization.problems import Variable
+
+try:
+    import cvxpy as cvx
+    _HAS_CVXPY = True
+except ImportError:
+    _HAS_CVXPY = False
+
+
+class GoemansWilliamsonOptimizationResult(OptimizationResult):
```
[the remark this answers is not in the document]
adekusar-drl: Added all new classes, hopefully in the correct place/order.

qiskit_optimization/algorithms/goemans_williamson_optimizer.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    Contains results of the Goemans-Williamson algorithm. The properties ``x`` and ``fval`` contain
+    values of just one solution. Explore ``all_solution`` for all possible solutions.
```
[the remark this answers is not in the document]
adekusar-drl: This is an artifact from the previous design. Now it is `samples`.

qiskit_optimization/algorithms/goemans_williamson_optimizer.py
```diff
@@ -0,0 +1,254 @@
...
+            seed: A seed value for the random number generator.
+
+        Raises:
+            MissingOptionalLibraryError: CVXPY is not installed.
+        """
+        if not _HAS_CVXPY:
+            raise MissingOptionalLibraryError(
+                libname='CVXPY',
+                name='GoemansWilliamsonOptimizer',
+                pip_install='pip install qiskit-optimization[cvxpy]')
```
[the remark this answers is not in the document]
adekusar-drl: Done. Just copied from somewhere else long time ago.

qiskit_optimization/algorithms/goemans_williamson_optimizer.py
```diff
@@ -0,0 +1,254 @@
...
+            constraints.append(x[i, i] == 1)
+
+        # objective function
+        expr = cvx.sum(cvx.multiply(adj_matrix, (np.ones((num_vertices, num_vertices)) - x)))
+
+        # solve
+        problem = cvx.Problem(cvx.Maximize(expr), constraints)
+        problem.solve()
+
+        # todo: add checks that the problem is solved
```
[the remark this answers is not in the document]
adekusar-drl: Actually yes, just added a check that we have a solution.

qiskit_optimization/algorithms/goemans_williamson_optimizer.py
```diff
@@ -0,0 +1,254 @@
...
+        # todo: weird numbers: 1.001 and 0.00001
+        if min(eigenvalues) < 0:
+            chi = chi + (1.001 * abs(min(eigenvalues)) * np.identity(num_vertices))
+        elif min(eigenvalues) == 0:
+            chi = chi + 0.00001 * np.identity(num_vertices)
+        x = np.linalg.cholesky(chi).T
+
+        r = np.random.normal(size=(self._num_cuts, num_vertices))
+
+        # todo: why "+ 0" ?
```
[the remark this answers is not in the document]
adekusar-drl: Removed.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
adekusar-drl: Removed, added a reference to the paper in the optimizer.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
adekusar-drl: Added.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
adekusar-drl: Removed.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
adekusar-drl: Removed, updated.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
adekusar-drl: Agree. Renamed.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
adekusar-drl: Added a check. Yes, it should be OptimizationError.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
adekusar-drl: Yes, we expect the one from Terra.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
adekusar-drl: Included the actual value.

setup.py
adekusar-drl: This was copied from Aqua as is. Honestly, no idea why it is so.

qiskit_optimization/algorithms/goemans_williamson_optimizer.py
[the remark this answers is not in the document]
woodsp-ibm: Seems fine thx. We will need to review the documentation ahead of release anyway. Given that the folders like algorithms etc are now toplevel the main page ought to be altered to reflect/enumerate these directly rather than via submodules. If there is some better grouping of the algorithms in the documentation that makes more sense for users that can be done then too.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
```diff
@@ -0,0 +1,345 @@
...
+
+    """
+
+    def __init__(self,
+                 pre_solver: OptimizationAlgorithm,
+                 relax_for_pre_solver: bool,
+                 qaoa: QAOA,
+                 epsilon: float = 0.25,
+                 num_initial_solutions: int = 1,
+                 circuit_factory: Optional[WarmStartQAOAFactory] = None,
```
woodsp-ibm, line 202: Given the class name changed would a change of parameter name here help too?

setup.py
woodsp-ibm: @manoelmarques I believe this set of versions came from elsewhere in qiskit which was using it in the past. The exclusions, as far as I recall, are because of problems with the set of function that was used by that code in those releases. The list was added to aqua to be consistent with versions supported. Ignis simply states >= 1.0.15 for a version here since presumably some function was added/changed then and that as a min is needed. If there is no specific version requirement for the functionality of cvxpy used here we could drop any version list or just use the same as Ignis since any install that included Ignis would need at least that level anyway.

manoelmarques: I removed the version list on cvxpy and added cvxpy install on CI.

qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py
adekusar-drl: This is a reasonable question. Since `circuit_factory` sort of a prohibited name, I suggest `warming_factory`.

woodsp-ibm: Would 'factory' be simple enough to do the job - the type is right there for further info/

adekusar-drl: I think `factory` is too generic, considering how often this term is used. So, I prefer have slightly more information in a variable name rather than just plain factory. This is why I suggest to have this name.

woodsp-ibm: How about `warm_start_factory`? The factory has does have more than just circuits, and maybe the meaning in the name here is closer to its use... Beyond that I am running out of ideas!

adekusar-drl: OK, let it be `warm_start_factory`.

woodsp-ibm (approved):

PR #22, opened 2021-02-04
Fix for docplex 2.20

### LINKED ISSUE #21: New version of Docplex breaks QuadraticProgram
### Information

- **Python version**: 3.8.7
- **Operating system**: macOS 11.2

The latest version of Docplex breaks `QuadraticProgram`
https://pypi.org/project/docplex/2.20.204/

```
$ python -m unittest test/test_quadratic_program.py
E
======================================================================
ERROR: test_quadratic_program (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_quadratic_program
  File "/usr/local/Cellar/python@3.8/3.8.7_2/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "/Users/ima/tasks/1_2021/qiskit/opt/test/test_quadratic_program.py", line 23, in <module>
    from qiskit_optimization import INFINITY, QiskitOptimizationError, QuadraticProgram
  File "/Users/ima/tasks/1_2021/qiskit/opt/qiskit_optimization/__init__.py", line 88, in <module>
    from .problems.quadratic_program import QuadraticProgram
  File "/Users/ima/tasks/1_2021/qiskit/opt/qiskit_optimization/problems/__init__.py", line 54, in <module>
    from .quadratic_program import QuadraticProgram
  File "/Users/ima/tasks/1_2021/qiskit/opt/qiskit_optimization/problems/quadratic_program.py", line 27, in <module>
    from docplex.mp.linear import Var
ImportError: cannot import name 'Var' from 'docplex.mp.linear' (/Users/ima/envs/dev38/lib/python3.8/site-packages/docplex/mp/linear.py)


----------------------------------------------------------------------
Ran 1 test in 0.000s

FAILED (errors=1)
```

The unit test passes

### Suggested solutions

I will fix it.

t-imamichi:
### Summary

Fixes #21

I tested both docplex 2.20.204 and docplex 2.19.202 with test_quadratic_program.py

### DIFF
--- a/qiskit_optimization/applications/ising/docplex.py
+++ b/qiskit_optimization/applications/ising/docplex.py
@@ -64,7 +64,7 @@ from math import fsum
 import numpy as np
 from docplex.mp.constants import ComparisonType
 from docplex.mp.constr import LinearConstraint, QuadraticConstraint
-from docplex.mp.linear import Var
+from docplex.mp.dvar import Var
 from docplex.mp.model import Model
 from qiskit.quantum_info import Pauli
 from qiskit.opflow import PauliSumOp
--- a/qiskit_optimization/problems/quadratic_program.py
+++ b/qiskit_optimization/problems/quadratic_program.py
@@ -24,7 +24,7 @@ import numpy as np
 from docplex.mp.constr import LinearConstraint as DocplexLinearConstraint
 from docplex.mp.constr import NotEqualConstraint
 from docplex.mp.constr import QuadraticConstraint as DocplexQuadraticConstraint
-from docplex.mp.linear import Var
+from docplex.mp.dvar import Var
 from docplex.mp.model import Model
 from docplex.mp.model_reader import ModelReader
 from docplex.mp.quad import QuadExpr
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ qiskit-terra>=0.17.0
 scipy>=1.4
 numpy>=1.17
 psutil>=5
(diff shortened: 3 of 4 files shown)


woodsp-ibm:
@t-imamichi Is the Var part of the public API of docplex or some internal class. @adekusar-drl went through a while back to change anything over to using public API (though its not so obvious in some cases) so we could try and avoid breaking any time a new docplex release came out if things had changed. This is of course assuming public API is deprecated etc so one has time to change things. Was it public and if so was it deprecated and we missed it - or was it another breaking change that we had been hoping to avoid going forwards via discussion with the docplex team.

t-imamichi:
I checked several versions of docplex. `mp.linear.Var` seems like an internal alias of `mp.dvar.Var`, which is removed in the latest version. Perhaps, we need to double-check the Docplex API.

woodsp-ibm:
> Perhaps, we need to double-check the Docplex API.

Perhaps we could also reach out to the docplex team to maybe scan through our use of docplex within our code to help out here too. I recall some things we had been looking at not being so obvious from the code/docs as to whether it was really part of what was considered an internal API used within the codebase versus a public API. I would imagine someone from their side looking over what we have would be able to indicate more easily if something were doing was using what was considered internal logic and that there was a better way to do things that we could avoid seeing these breakages.

manoelmarques (approved):

PR #23, opened 2021-02-05
Deal with both old and new locations of Var of docplex

t-imamichi:
### Summary

Follow-up of #22 

I replaced `mp.linear.Var` with `mp.dvar.Var` to be compatible with the latest docplex 2.20.204.
`mp.linear.Var` was used until 2.15.194. So, I modified the code to work with both styles.

I add a version pin docplex==2.15.194 for macOS because docplex does not work without cplex on macOS.
See #24 for details.

### DIFF
--- a/qiskit_optimization/applications/ising/docplex.py
+++ b/qiskit_optimization/applications/ising/docplex.py
@@ -57,17 +57,24 @@ The following is an example of use.
 
 """
 
-from typing import Tuple, Union, List
 import logging
 from math import fsum
+from typing import Tuple, Union, List
 
 import numpy as np
 from docplex.mp.constants import ComparisonType
 from docplex.mp.constr import LinearConstraint, QuadraticConstraint
-from docplex.mp.dvar import Var
+
+try:
+    # new location since docplex 2.16.196
+    from docplex.mp.dvar import Var
+except ImportError:
+    # old location until docplex 2.15.194
+    from docplex.mp.linear import Var
 from docplex.mp.model import Model
-from qiskit.quantum_info import Pauli
+
 from qiskit.opflow import PauliSumOp
+from qiskit.quantum_info import Pauli
 
 from ...exceptions import QiskitOptimizationError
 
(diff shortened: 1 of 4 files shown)


manoelmarques (dismissed):

woodsp-ibm:
I see 2.16.196 was released in early Nov last year - so its not that old. If the two classes are compatible in terms of internals could we do code that imported the Var from the new location and if that failed try the location where we imported it from before? This would give a broader compatibility if its reasonably doable.

t-imamichi:
It can be realized as follows. What do you think?
```python
try:
  # new location for 2.16.196 (Nov 2020) or newer
  from docplex.mp.dvar import Var
except ImportError:
  # old location for 2.15.194 (Jul 2020) or before
  from docplex.mp.linear import Var
```

t-imamichi:
I convert this to a draft to discuss.

woodsp-ibm:
That seems great. I looked and see it would affect 2 files. Here in optimization is there are reason why we want to keep the docplex ising convertor in applications? If that vanished then it would just be the one place for that very simple change in QuadraticeProgram (in Aqua we need it the 2 files still as that general docplex ising application would have to stay for compatibility as the whole lot is deprecated)

t-imamichi:
I think docplex ising converter will be removed. But, I will change both two files just in case.

t-imamichi:
I remember that @stefan-woerner said that qiskit-optimization supports only new location. Is it correct? If so, I close this issue.

woodsp-ibm:
In regards of compatibilty an pinning take a look at comments added today to #24 

For Var locations I was arguing that for some institutions that their installs may be more regulated and since the code had worked with the older docplex versions requiring the latest version (or some specific one) may be problematic. Since the fix is really just a conditional import in one place, ie. QuadraticProgram, it seems that it would be valuable to have it that way (I ignored the docplex ising convertor as I am assuming that would be deleted from here). With the code being able to handle either Var location, via the try/catch on the import, it would not preclude us from limiting versions in setup.py if that were needed too.

t-imamichi:
OK. Then, I keep the conditional import to deal with both old and new locations. As I commented to #24, the docplex issue without cplex occurs only on macOS. So, how about the following requirements.txt?

```
docplex; sys_platform != "darwin"
docplex==2.15.196; sys_platform == "darwin"
```

woodsp-ibm:
Ok, lets go with this for now then. Strange that the issue is only on MacOS, wonder what the difference is on that platform in the way the optional cplex is handled by docplex.

woodsp-ibm (approved):

PR #31, opened 2021-02-15
Redesign application classes for optimization problems

a-matsuo:
### Summary
**This is a WIP PR. Any comments are more than welcome. Especially, I don't think that the function names are the best ones.**

This PR is for application classes for optimization problems such as maxcut and tsp. These classes are for application users who do not have deep knowledges of quantum computers and/or quadratic programing. They can directly recieve a graph as an input and can create a quadratic program model automatically (The type of input depends on an application). Also, they can "interpret" a binary solution [1, 0, 0,...,1] and can plot a solution graph based on it. These classes are redesign of old files for optimization problems located under `qiskit-optimization/application/ising`. This is a usage [example for the 'Maxcut' class](https://github.com/a-matsuo/qiskit-optimization/blob/optimization_problems/qiskit_optimization/applications/examples/maxcut_example.ipynb)

[new_application_class.pdf](https://github.com/Qiskit/qiskit-optimization/files/5981004/new_application_class.pdf)

problems

- [x] BaseApplication
- [x] GraphApplication
- [x] Clique ([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/clique_example.ipynb))
- [x] GraphPartitioning ([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/graph_partitioning_example.ipynb))
- [x] Maxcut ([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/maxcut_example.ipynb))
- [x] StableSet([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/stable_set_example.ipynb))
- [x] Tsp ([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/tsp_example.ipynb))
- [x] VehicleRouting([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/vehicle_routing_example.ipynb))
- [x] VertexCover([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/vertex_cover_example.ipynb))
- [x] ExactCover([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/exact_cover_example.ipynb))
- [x] KNapsack([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/knapsack_example.ipynb))
- [x] NumberPartinioning([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/number_partitioning_example.ipynb))
- [x] SetPacking([example](https://github.com/a-matsuo/optimization_class/blob/main/examples/set_packing.ipynb))


t-imamichi:
Because this is not yet completed, it is better to use "draft" PR. I converted this to "draft".

t-imamichi:
jupyter notebooks won't be part of qiskit optimization. It would be better to move them to another repo or gist.

qiskit_optimization/applications/ising/base_application.py
```diff
@@ -0,0 +1,25 @@
...
+        raise NotImplementedError
+
+    def is_feasible(self, x):
+        return self._build_quadratic_program().is_feasible(x)
+
+    def evaluate(self, x):
+        return self._build_quadratic_program().objective.evaluate(x)
+
+    def to_quadratic_program(self):
+        return copy.deepcopy(self._qp)
```
stefan-woerner, line 25: Why use `self._qp` here but not in `evaluate` and `is_feasible` ? If we store it, we need to make sure we delete it or re-build it in case anything changes in an application object.

qiskit_optimization/applications/ising/graph_application.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def __init__(self, g):
+        self._g = g.copy(as_view=True)
+        self._qp = self._build_quadratic_program()
+
+    @abstractmethod
+    def plot_graph(self):
+        raise NotImplementedError
+
+    def g(self):
```
stefan-woerner, line 21: please rename to `graph`

qiskit_optimization/applications/ising/graph_application.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    An abstract class for graph problems in optimization problems
+    """
+
+    def __init__(self, g):
+        self._g = g.copy(as_view=True)
+        self._qp = self._build_quadratic_program()
+
+    @abstractmethod
+    def plot_graph(self):
```
stefan-woerner, line 18: should optionally allow a solution vector `x`.

qiskit_optimization/applications/ising/base_application.py
woodsp-ibm: If we want to have a BaseApplication class here would it make sense to somehow separate the problem part of the application, that perhaps can be treated more generically, from the aspect of the quadratic_program that may not be the only path in the future to solving the problem?

In respect of this class on self-._qp it refers to a variable that is not part of this class. Maybe there should be some private method to set it and/or pass it up to a constructor here to store it. Maybe it supposed to be stored/cached via the build_quadratic_program, in some manner.

qiskit_optimization/applications/ising/graph_application.py
woodsp-ibm: Maybe consider using `draw` here as circuits can be rendered using draw, so this would make things consistent, and this is what nx has for drawing them too. Then the result can hopefully be *drawn* not matter whether its a set or a graph etc.

PS. Don't forget to include the standard copyright text in the content here.

qiskit_optimization/applications/ising/max_cut.py
```diff
@@ -1,83 +1,38 @@
...
+        for u, v in self._g.edges:
+            self._g.edges[u, v].setdefault('weight', 1)
+        objective = mdl.sum(self._g.edges[i, j]['weight'] * x[i]
+                            * (1 - x[j]) for i, j in self._g.edges)
+        mdl.maximize(objective)
+        qp = QuadraticProgram()
+        qp.from_docplex(mdl)
+        return qp
+
+    def plot_graph(self, x, pos=None):
```
woodsp-ibm, line 27: `x` seems such a generic name when looking at this method. From the notebook sample this seems to be `x` from the result (or some other equivalent value of your choosing that may not be the result). Would it make sense to be able to pass the Result object here or `x` such that its easy to pass the result, which seems to be more the usual use case to me such that x is taken from it, as well as passing one more directly to try other values.

qiskit_optimization/applications/ising/max_cut_old.py
```diff
@@ -0,0 +1,83 @@
+# This code is part of Qiskit.
```
woodsp-ibm, line 1: I assume this 'old' file will just be deleted? (Same for the docplex old one)

Now it may be educational to have some direct to ising mapping in the maxcut app as its easy to follow. Of course there are papers out there about solving these type of problems with this mechanism so maybe it can be just simply removed, as you can get an ising from the QP, though its more work to follow/understand that path.

qiskit_optimization/applications/ising/tsp.py
```diff
@@ -1,274 +1,110 @@
...
+    def interpret(self, x):
+        n = int(np.sqrt(len(x)))
+        route = []
+        for p__ in range(n):
+            for i in range(n):
+                if x[i * n + p__]:
+                    route.append(i)
+        return route
+
+    @staticmethod
```
woodsp-ibm, line 51: Would this method and the parse_tsplib_format method not be better creating a TSP object with the graph so the user directly gets the TSP application rather than a graph that they then have to turn into a TSP to use at this application level.

qiskit_optimization/applications/ising/tsp.py
```diff
@@ -1,274 +1,110 @@
...
+            coord_section = False
+            for line in infile:
+                if line.startswith('NAME'):
+                    name = line.split(':')[1]
+                    name.strip()
+                elif line.startswith('TYPE'):
+                    typ = line.split(':')[1]
+                    typ.strip()
+                    if typ != 'TSP':
+                        logger.warning('This supports only "TSP" type. Actual: %s', typ)
```
woodsp-ibm, line 84: I am thinking that perhaps this logger.warning (and the one below) ought to be raising a QiskitOptimizationError rather than logging a warning and continuing. The warning may be missed and it seems in these cases that the file contains something that is not supported by the TSP application here so would be better as an error?

qiskit_optimization/applications/ising/base_application.py
woodsp-ibm: As the class is now being used in another domain, I know naming may not be sorted yet, but perhaps when it is something like OptimizationApplication may be a more apt base class name that when looking at a FinanceApp, that is derived, its more obvious than a generically named BaseApplication

a-matsuo:
@t-imamichi @stefan-woerner @woodsp-ibm 
Thank you so much for you guys' helpful comments. I'm fixing them!

qiskit_optimization/applications/ising/max_cut_old.py
a-matsuo: I was thinking to delete them, but we haven't discussed about existing files, so I just renamed and kept them. If there is some good place, we can move them to there (or we can make some tutorial based on existing files in community-tutorial? I don't know).

a-matsuo:
I made working codes for all the optimization application classes. I haven't added unittest and comments yet, so I'm going to add them and finalize codes now.

qiskit_optimization/applications/ising/base_application.py
```diff
@@ -0,0 +1,23 @@
+import copy
+from abc import ABC, abstractmethod
+
+
+class BaseApplication(ABC):
```
stefan-woerner, line 5: Let's rename to `BaseOptimizationApplication` so we can also have different ones.

qiskit_optimization/applications/ising/graph_application.py
```diff
@@ -0,0 +1,27 @@
+from typing import Optional
+from abc import abstractmethod
+
+import networkx as nx
+
+from .base_application import BaseApplication
+
+
+class GraphApplication(BaseApplication):
```
stefan-woerner, line 9: as in previous comment, maybe more specific `GraphOptimizationApplication` would be better.

qiskit_optimization/applications/ising/max_cut.py
```diff
@@ -1,83 +1,42 @@
...
+        mdl.maximize(objective)
+        qp = QuadraticProgram()
+        qp.from_docplex(mdl)
+        return qp
+
+    def draw_graph(self, result=None, pos=None):
+        if result is None:
+            nx.draw(self._graph, pos=pos, with_labels=True)
+        else:
+            colors = ['r' if value == 0 else 'b' for value in result.x]
```
stefan-woerner, line 29: use `_node_colors`?

qiskit_optimization/applications/ising/max_cut.py
```diff
@@ -1,83 +1,42 @@
...
+            nx.draw(self._graph, node_color=colors, pos=pos, with_labels=True)
```
stefan-woerner, line 30: use`pos` either everywhere (i.e. in the parent class) or in none of the graph apps.

stefan-woerner (changes requested):

qiskit_optimization/applications/ising/knapsack.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, values, weights, max_weight):
         self._max_weight = copy.deepcopy(max_weight)
 
     def to_quadratic_program(self):
-        mdl = Model(name='K_napsack')
+        mdl = Model(name='K napsack')
```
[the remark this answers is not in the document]
a-matsuo: Ah, I was wondering why it's called "K napsack" in English like K means, but we don't pronounce K in knapsack😉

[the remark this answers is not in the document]
woodsp-ibm: Right - K is not pronounced - same as you would do for Knock

a-matsuo:
I've finished writing codes including unittest, comments, and typhints. What shall I do with old application files? I temporary moved them into the `existing` folder.

stefan-woerner:
@a-matsuo great, you can delete all the old files (they still exist in qiskit-aqua in case we need to look at them).

qiskit_optimization/applications/ising/vehicle_routing.py
```diff
@@ -10,187 +10,215 @@
...
+
+    def to_quadratic_program(self) -> QuadraticProgram:
+        """Convert a vehicle routing problem instance into a
+        :class:`~qiskit_optimization.problems.QuadraticProgram`
+
+        Returns:
+            The :class:`~qiskit_optimization.problems.QuadraticProgram` created
+            from the vehicle routing problem instance.
+        """
+        mdl = Model(name='Vehicle Routing')
```
woodsp-ibm, line 58: I note in this name both words are capitalized, whereas it was `Stable set` above for that one, where only the first work was capitalized. Its a minor comment, I just noticed the difference that was all.

qiskit_optimization/applications/ising/base_optimization_application.py
```diff
@@ -0,0 +1,36 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+from abc import ABC, abstractmethod
+
+from qiskit_optimization.algorithms import OptimizationResult
+
+
+class BaseOptimizationApplication(ABC):
```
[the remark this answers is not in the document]
stefan-woerner: I like `OptimizationApplication`. We can get rid of the `ising` folder. For `qiskit_finance` we should then also restructure to `applications.optimization` and `applications.estimation`. In case we have more problem classes than QPs at a later stage, we might want to introduce some more structure, but I agree, right now, the `ising` doesn't help much.

qiskit_optimization/applications/ising/clique.py
```diff
@@ -10,135 +10,110 @@
...
+        Returns:
+            The list of node indices whose corresponding variable is 1
+        """
+        clique = []
+        for i, value in enumerate(result.x):
+            if value:
+                clique.append(i)
+        return clique
+
+    def draw_graph(self, result: Optional[OptimizationResult] = None,
```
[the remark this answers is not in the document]
stefan-woerner: I opt for `draw`

qiskit_optimization/applications/ising/existing/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+
+"""
+Ising Models (:mod:`qiskit_optimization.applications.ising`)
+============================================================
+
+Ising models for optimization application problems
+
+.. currentmodule:: qiskit_optimization.applications.ising
+
+Ising Models
```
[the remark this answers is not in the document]
stefan-woerner: let's just have `qiskit_optimization.applications` for now and restructure as needed when we have more information.

a-matsuo:
Thank you so much for the helpful comments. I finished fixing all of them.

stefan-woerner (dismissed):

qiskit_optimization/applications/knapsack.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, values: List[int], weights: List[int], max_weight: int) -> None:
+        """
+        Args:
+            values: A list of the values of items
+            weights: A list of the weights of items
+            max_weight: The maximum weight capacity
+        """
+        self._values = values
+        self._weights = weights
+        self._max_weight = max_weight
```
woodsp-ibm, line 41: I see some applications allow aspects to be changed. Like in clique `size` is changeable, and in vehicle routine `num_vehicles` and `depot`  are.  It seems `max_weight` could be settable the same way, so given the same set of items, try the same problem with different max weights, like trying vehicle routing with a different number of vehicles...

test/applications/test_clique.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_interpret(self):
+        """Test interpret"""
+        clique = Clique(self.graph)
+        self.assertEqual(clique.interpret(self.result), [0, 2, 3, 4])
+
+    def test_node_colors(self):
+        """Test _node_colors"""
+        clique = Clique(self.graph)
+        self.assertEqual(clique._node_colors(self.result), ['r', 'darkgrey', 'r', 'r', 'r'])
```
woodsp-ibm, line 73: I guess this is because there is no reasonable way to test the public 'draw' API? Ideally we should be testing these internal private methods via the public API of the class, not doing so directly like this

test/applications/test_clique.py
```diff
@@ -0,0 +1,73 @@
...
+import networkx as nx
+
+from qiskit_optimization import QuadraticProgram
+from qiskit_optimization.algorithms import (OptimizationResult,
+                                            OptimizationResultStatus)
+from qiskit_optimization.applications.clique import Clique
+from qiskit_optimization.problems import (Constraint, QuadraticObjective, VarType)
+
+
+class TestClique(QiskitOptimizationTestCase):
```
woodsp-ibm, line 25: Since `size` is a public method (getter/setter) there should be unit tests that validate it works as expected etc.

test/applications/test_clique.py
```diff
@@ -0,0 +1,73 @@
...
+        for _ in range(5):
+            op.binary_var()
+        self.result = OptimizationResult(
+            x=[1, 0, 1, 1, 1], fval=4, variables=op.variables,
+            status=OptimizationResultStatus.SUCCESS)
+
+    def test_to_quadratic_program(self):
+        """Test to_quadratic_program"""
+        clique = Clique(self.graph)
+        op = clique.to_quadratic_program()
```
woodsp-ibm, line 42: I guess we have any equality method on Quadratic Program such that some reference QP could be generated here and then one simple test for equality made? I see there are similar logic in other tests  to test the QP is as expected. Maybe some for of equals would be useful? If not in QP itself then maybe as some util function, perhaps in the tests here even, that would help avoid doing this test explicitly piece by piece each time we need it.

qiskit_optimization/applications/clique.py
```diff
@@ -0,0 +1,120 @@
...
+    def size(self) -> int:
+        """Getter of size
+
+        Returns:
+            The size of the clique
+        """
+        return self._size
+
+    @size.setter
+    def size(self, size: int) -> None:
```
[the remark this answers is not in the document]
a-matsuo: It was difficult to understand without comments. I added some comments. Changing the value of `size` does not affect interpreting a result or drawing the result. When "size" is None, `to_quadratic_probgram` makes an optimization model for a maximal clique instead of the specified size of a clique.

test/applications/test_clique.py
a-matsuo: Yes, I could not find a reasonable way to test Networkx's `draw` with my own coloring method, so I just tested `_node_colors` which is a private method. I will make a test via the public API of the class for other cases.
```
nx.draw(self._graph, node_color=self._node_colors(x), pos=pos, with_labels=True)
```

test/applications/test_clique.py
t-imamichi: No, there is no equality method of `QuadraticProgram`. It is generally used only for unit tests. So, I recommend to add an equality function to a unit test.

a-matsuo: Somehow `QuadraticProgram` does not have a method for equality. It test explicitly even in the test of `QuadraticProgram`. I didn't think about it, but, you're right. I could make some helper function like `assertQPEqual` for those tests. I will use a helper function next time as you taught me.

qiskit_optimization/applications/graph_optimization_application.py
```diff
@@ -0,0 +1,116 @@
...
+        """An abstract method to draw the graph based on the result.
+
+        Args:
+            result: The calculated result for the problem
+            pos: The positions of nodes
+
+        Raises:
+            NotImplementedError: When this function is not implemented in sub classes.
+        """
+        pass
```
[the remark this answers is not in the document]
woodsp-ibm: if each sub-class had a _draw_result() it would be even clearer I think. Base class can draw the problem, but it needs the  derived class to draw the result. I just thought it might be easier if we ever added more that each sub-class could then not have to worry about drawing the problem. But its not a needed change as such since the code would work as-is. Given whatever else you are doing I'll leave it to you as to whether its something you think is worthwhile doing now and have the time etc

test/applications/test_clique.py
woodsp-ibm: Yes, in this case having test is better than not having one there even though this is only an aspect of the draw method.

qiskit_optimization/applications/graph_optimization_application.py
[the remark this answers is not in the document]
a-matsuo: Thank you for your comments. Ya, you're right. I added `_draw_result` for future maintenance 👍

woodsp-ibm (approved):

PR #33, opened 2021-02-19
Change `GroverOptimizerResult` to have `samples`

a-matsuo:
### Summary
Now `GroverOptimizerResult` has information of `samples` like `MinimumEigenOptimizationResult` does.

### Details and comments
**Note that this PR needs to be merged after https://github.com/Qiskit/qiskit-terra/pull/5870 is merged.**


a-matsuo:
I found this line is causing some issue. When I printed out feq[-1] before and after the line. The result became as follows.
```
before ('10101110110', 6e-05)
after ('10101110110', 0.0007000000000028095)
before ('00001111110', 6e-05)
after ('00001111110', 0.0007000000000028095)
before ('10101111110', 6e-05)
after ('10101111110', 0.0007000000000028095)
before ('11101111110', 4e-05)
after ('11101111110', -5.9999999998616715e-05)
```
In the last line, the value `4e-05` of `11101111110` became `-5.9999999998616715e-05` which is a negative number because of the calculation error with small float numbers. This causes ValuError in Numpy

_generator.pyx in numpy.random._generator.Generator.choice()
ValueError: probabilities are not non-negative
I’m not sure why we originally have this line here. It’s basically re-calculating the value `f[-1]` by `1.0 - sum(x[1] for x in freq[0:len(freq) - 1]`. I think we can remove it, and then the code will work without this issue.

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -278,10 +292,9 @@ def _measure(self, circuit: QuantumCircuit) -> str:
         probs = self._get_probs(circuit)
         freq = sorted(probs.items(), key=lambda x: x[1], reverse=True)
         # Pick a random outcome.
-        freq[-1] = (freq[-1][0], 1.0 - sum(x[1] for x in freq[0:len(freq) - 1]))
+        freq[-1] = (freq[-1][0], 1.0 - math.fsum(x[1] for x in freq[0:len(freq) - 1]))
```
t-imamichi, line 295: I don't think this line is necessary any more because of the probability normalization in _get_probs. Could you check it?

t-imamichi:
```suggestion
```

a-matsuo: I checked both of both qasm_simulator and statevector_simulator. It raises no errors without the line now. I think we don't need the line anymore since we normalized the `probs`, so I deleted the line.

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -335,8 +353,14 @@ def __init__(self, x: Union[List[float], np.ndarray], fval: float, variables: Li
                 that is expected to be identical with ``fval``.
             threshold: The threshold of Grover algorithm.
             status: the termination status of the optimization algorithm.
+            samples: the x value, the objective function value of the original problem,
+                the probability, and the status of sampling.
+            raw_samples: the x values of the QUBO, the objective function value of the QUBO,
```
woodsp-ibm, line 358: `samples` says its the `x value` (not 's' so singular), whereas `raw_samples` says `x values` i.e. plural. Are `samples` and `raw_samples` the same type so that 'raw_samples' should say `x value` just like `samples' does. They both seem to be a List[SolutionSample] so having one as 'x value' and the other as 'x values' seems misleading - or it did to me as the latter seems to be more than one x value, which samples is not.

qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -444,3 +445,94 @@ def _interpret(cls, x: np.ndarray,
                             variables=problem.variables,
                             status=cls._get_feasibility_status(problem, x),
                             **kwargs)
+
+    def _interpret_samples(self, problem: QuadraticProgram, raw_samples: List[SolutionSample],
```
t-imamichi, line 449: Because `_get_feasibility_status` is a static method, you should make it `@classmethod`.

qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -444,3 +445,94 @@ def _interpret(cls, x: np.ndarray,
...
+        for key, x in array.items():
+            probability = prob[key]
+            fval = problem.objective.evaluate(x)
+            status = self._get_feasibility_status(problem, x)
+            samples.append(SolutionSample(x, fval, probability, status))
+
+        return sorted(samples,
+                      key=lambda v: (v.status.value, problem.objective.sense.value * v.fval))
+
+    def _eigenvector_to_solutions(self, eigenvector: Union[dict, np.ndarray, StateFn],
```
t-imamichi, line 471: I don't see any usage of `self`. You should make it `@staticmethod`.

qiskit_optimization/algorithms/grover_optimizer.py
a-matsuo: Right, I forgot to put "s" for x value in `samples`. Thanks!

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -17,16 +17,18 @@
 from copy import deepcopy
 from typing import Optional, Dict, Union, List, cast
 
+
```
t-imamichi, line 20:
```suggestion
```

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -278,28 +292,30 @@ def _measure(self, circuit: QuantumCircuit) -> str:
...
         """Gets probabilities from a given backend."""
         # Execute job and filter results.
         result = self.quantum_instance.execute(qc)
         if self.quantum_instance.is_statevector:
-            state = np.round(result.get_statevector(qc), 5)
+            state = result.get_statevector(qc)
             keys = [bin(i)[2::].rjust(int(np.log2(len(state))), '0')[::-1]
                     for i in range(0, len(state))]
-            probs = [np.round(abs(a) * abs(a), 5) for a in state]
+            probs = [abs(a) * abs(a) for a in state]
```
t-imamichi, line 307: If `state` is ndarray, we can write it in a simplified way. If it is a list, you can simplify as follows.
```suggestion
            probs = [abs(a) ** 2 for a in state]
```

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -278,28 +292,30 @@ def _measure(self, circuit: QuantumCircuit) -> str:
...
             hist = dict(zip(keys, probs))
+            self._circuit_results = state
         else:
             state = result.get_counts(qc)
             shots = self.quantum_instance.run_config.shots
             hist = {}
             for key in state:
                 hist[key[::-1]] = state[key] / shots
+            self._circuit_results = {b[::-1]: (v / shots) ** 0.5 for (b, v) in state.items()}
         hist = dict(filter(lambda p: p[1] > 0, hist.items()))
```
t-imamichi, line 319: You can eliminate this line by removing items with value 0 when `hist` is generated appropriately.

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -278,28 +292,30 @@ def _measure(self, circuit: QuantumCircuit) -> str:
...
         if self.quantum_instance.is_statevector:
-            state = np.round(result.get_statevector(qc), 5)
+            state = result.get_statevector(qc)
             keys = [bin(i)[2::].rjust(int(np.log2(len(state))), '0')[::-1]
                     for i in range(0, len(state))]
-            probs = [np.round(abs(a) * abs(a), 5) for a in state]
+            probs = [abs(a) * abs(a) for a in state]
+            total = math.fsum(probs)
+            probs = [p / total for p in probs]
             hist = dict(zip(keys, probs))
```
t-imamichi, line 310:
```suggestion
            hist = {key: prob for key, prob in zip(keys, probs) if prob > 0}
```

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -278,28 +292,30 @@ def _measure(self, circuit: QuantumCircuit) -> str:
...
+            self._circuit_results = state
         else:
             state = result.get_counts(qc)
             shots = self.quantum_instance.run_config.shots
             hist = {}
             for key in state:
                 hist[key[::-1]] = state[key] / shots
```
t-imamichi, line 317:
```suggestion
            hist = {key[::-1]: val / shots for key, val in state.items() if val > 0}
```

t-imamichi (approved):

a-matsuo:
Thank you so much for reviewing! @t-imamichi @woodsp-ibm,
I've finished fixing all the commented points. The PR should be ready now.

stefan-woerner (approved):

woodsp-ibm (approved):

PR #35, opened 2021-02-25
Updated recursive optimizer

adekusar-drl:
### Summary
Since we have introduced `samples` property in the `OptimizationResult` we can make `RecursiveMinimumEigenOptimizer` more generic and accept wider range of optimizers. This is what is done in this PR.

### DIFF
--- a/qiskit_optimization/algorithms/recursive_minimum_eigen_optimizer.py
+++ b/qiskit_optimization/algorithms/recursive_minimum_eigen_optimizer.py
@@ -99,7 +99,9 @@ class RecursiveMinimumEigenOptimizer(OptimizationAlgorithm):
     """A meta-algorithm that applies a recursive optimization.
 
     The recursive minimum eigen optimizer applies a recursive optimization on top of
-    :class:`~qiskit_optimization.algorithms.MinimumEigenOptimizer`.
+    :class:`~qiskit_optimization.algorithms.OptimizationAlgorithm`. This optimizer can use
+    :class:`~qiskit_optimization.algorithms.MinimumEigenOptimizer` as an optimizer that is called
+    at each iteration.
     The algorithm is introduced in [1].
 
     Examples:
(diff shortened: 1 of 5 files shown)


stefan-woerner (dismissed):
lgtm

adekusar-drl:
@stefan-woerner thanks, I'd like to see if @woodsp-ibm has any comments.

qiskit_optimization/algorithms/recursive_minimum_eigen_optimizer.py
```diff
@@ -135,7 +135,7 @@ def __init__(self, min_eigen_optimizer: MinimumEigenOptimizer, min_num_vars: int
         variables is reached.
 
         Args:
-            min_eigen_optimizer: The eigen optimizer to use in every iteration.
+            optimizer: The optimizer to use in every iteration.
             min_num_vars: The minimum number of variables to apply the recursive scheme. If this
                 threshold is reached, the min_num_vars_optimizer is used.
             min_num_vars_optimizer: This optimizer is used after the recursive scheme for the
```
woodsp-ibm, line 141: I know this was not changed, but perhaps we should state in the docstring that internally this defaults to the MinEigOpt.. when None.

qiskit_optimization/algorithms/recursive_minimum_eigen_optimizer.py
```diff
@@ -203,11 +203,11 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
 
         # run recursive optimization until the resulting problem is small enough
         replacements = {}  # type: Dict[str, Tuple[str, int]]
-        min_eigen_results = []  # type: List[MinimumEigenOptimizationResult]
+        min_eigen_results = []  # type: List[OptimizationResult]
```
woodsp-ibm, line 206: Since this is now a more general optimizer would the name 'min_eigen_results' be "better" say as a more general name say 'optimization_results' or something - that may be better for any readers of the code in future. I don't think the name is externalized in any result as such - seems just used to form the history tuple later.

qiskit_optimization/algorithms/recursive_minimum_eigen_optimizer.py
adekusar-drl: Good point. Added. Also I reviewed the docstrings, there a few updates.

qiskit_optimization/algorithms/recursive_minimum_eigen_optimizer.py
adekusar-drl: Yes, updated.

woodsp-ibm (approved):

Issue #34, opened 2021-02-24
LinearEqualityToPenalty.penalty doesn't return the penalty factor used in conversion


peguerosdc:
### Information

- **Qiskit Optimization version**: latest
- **Python version**: 3.8.5
- **Operating system**: Ubuntu 20.04.2 LTS

### What is the current behavior?
After creating a converter `LinearEqualityToPenalty` with no penalty factor and converting a `QuadraticProgram`, calling `.penalty` doesn't return the penalty factor used in conversion as stated in the docs.

### Steps to reproduce the problem
```` python
# Create a quadratic program
qp = QuadraticProgram()
# After adding your constraints, convert the inequality constraints to penalties with LinearEqualityToPenalty
converter = LinearEqualityToPenalty()
converted_qp = converter.convert(qp)
# Calling .penalty returns None
converter.penalty
````

### What is the expected behavior?
Docs say that calling `.penalty` "returns the penalty factor used in conversion" and that is the expected result, but instead it returns `None` when no default penalty is set.

### Suggested solutions
Checking the code, when no penalty is provided in the constructor and `convert` is called, a penalty factor is calculated using `LinearEqualityToPenalty._auto_define_penalty()`, but `self._penalty` is not updated.

I can fix it as it would be just adding a single line after calculating the penalty in `convert` like this:

````python
if self._penalty is None:
    penalty = self._auto_define_penalty()
    self._penalty = penalty
else:
    penalty = self._penalty
````
(and maybe it's not much, but it would be my first Qiskit contribution ever 🎉), or if it is coded like this for a reason, I think the docs should be updated as they are a little misleading (I can do it as well if the rationale behind that decision is explained to me 🙂 )

a-matsuo:
Hi @peguerosdc 
thank you for the issue! Currently, `self._penalty` is used as a flag for whether the converter calls `self._auto_define_penalty()` or not. If we update `self._penalty` and fix it, `self._auto_define_penalty()` will not be called when people reuse the converter. How about adding a new variable something like `auto_define_penalty` variable, and we use it for the flag of `self._auto_define_penalty()` and update `self._penalty`? In this way, we can correctly reuse the converter. It might be a little bit difficult task for the first contribute. Can you do this?

PR #43, opened 2021-03-08
Retrieval of calculated penalty in LinearEqualityToPenalty converter

peguerosdc:
### Summary

`LinearEqualityToPenalty` converter now lets the user retrieve the last penalty factor calculated after a conversion when no default penalty was given. Please refer to issue (#34) for more details.

### Details and comments
A new `LinearEqualityToPenalty` property was defined called `_should_define_penalty` which remembers if a given penalty is valid (i.e. if a new penalty factor has to be calculated when calling `convert`).
This was previously handled by the variable `_penalty`, which is now updated on every conversion so the user can retrieve it by calling `.penalty`.

Documentation is updated on `LinearEqualityToPenalty` and `QuadraticProgramToQubo` as the latter makes use of the former.


qiskit_optimization/converters/linear_equality_to_penalty.py
```diff
@@ -59,8 +60,8 @@ def convert(self, problem: QuadraticProgram) -> QuadraticProgram:
         self._src = copy.deepcopy(problem)
         self._dst = QuadraticProgram(name=problem.name)
 
-        # If penalty is None, set the penalty coefficient by _auto_define_penalty()
-        if self._penalty is None:
+        # If no penalty was given, set the penalty coefficient by _auto_define_penalty()
+        if self._should_define_penalty:
```
a-matsuo, line 64: You need to update `self._penalty` after calling `_auto_define_penalty()`.

qiskit_optimization/converters/linear_equality_to_penalty.py
```diff
@@ -36,11 +36,12 @@ def __init__(self, penalty: Optional[float] = None) -> None:
...
         Args:
             penalty: Penalty factor to scale equality constraints that are added to objective.
-                     If None is passed, penalty factor will be automatically calculated.
+                     If None is passed, a penalty factor will be automatically calculated on
+                     every conversion.
         """
         self._src = None  # type: Optional[QuadraticProgram]
         self._dst = None  # type: Optional[QuadraticProgram]
-        self._penalty = penalty  # type: Optional[float]
+        self.penalty = penalty  # type: Optional[float]
```
a-matsuo, line 44: You don't need to change it to public.

a-matsuo (changes requested):
Could you please add `should_define_penalty` to the constructor of the converter? Also, could you add the setter/getter for that? So, users can choose whether they use `_auto_define_penalty()` or not.
`def __init__(self, penalty: Optional[float] = None) -> None:`

qiskit_optimization/converters/linear_equality_to_penalty.py
peguerosdc: But, wouldn't adding `should_define_penalty` to the constructor create conflict between using `penalty` and calculating a new one? I mean, if I do something like `LinearEqualityToPenalty(penalty=5, should_define_penalty=True)`, what would be the expected behavior?

qiskit_optimization/converters/linear_equality_to_penalty.py
peguerosdc: I am not changing it to public, I am using the setter when caling `self.penalty` (because the setter updates `_should_define_penalty`)

qiskit_optimization/converters/linear_equality_to_penalty.py
a-matsuo: You can make the default of `should_define_penalty` is `True`, and then if `penalty` is given explicitly the `penalty` is used instead of calling `_auto_define_penalty()`.

qiskit_optimization/converters/linear_equality_to_penalty.py
a-matsuo: I mean that shouldn't it be `self._penalty` instead of `self.penalty`.

t-imamichi:
If `LinearEqualityToPenalty` is used only for a single `QuadraticProgram`, it is realized by just storing the penalty values in `self._penalty`. But, if `LinearEqualityToPenalty` can be used for various `QuadraticProgram`, we should not store the penalty value to `self._penalty` because it cannot invoke `_auto_define_penalty` due to `self._penalty` value.

It would be good to add `self._use_auto_penalty: bool` to control the penalty value explicitly and store the penalty value to `self._penalty: Optional[float]` (if convert is never called, it may return `None`).

qiskit_optimization/converters/linear_equality_to_penalty.py
```diff
@@ -120,6 +121,9 @@ def convert(self, problem: QuadraticProgram) -> QuadraticProgram:
         else:
             self._dst.maximize(offset, linear, quadratic)
 
+        # Update the penalty to the one just used
+        self._penalty = penalty  # type: float
```
peguerosdc, line 125: @a-matsuo forgot to mention that I am updating `_penalty` in here instead of immediately after calling `_auto_define_penalty()` because I thought it would be better to update it after the conversion was made, so the value is not updated it something failed in the middle, though I can move it to line 64 as suggested :)

qiskit_optimization/converters/linear_equality_to_penalty.py
peguerosdc: I think I get what you are saying, but removing the `_` is not making the variable public (or creating a new one) because assigning a value to `penalty` (like `self.penalty = something`) is actually calling the setter as defined with `@penalty.setter`.

peguerosdc:
I think that exposing the inner logic of how `LinearEqualityToPenalty` decides whether to calculate the penalty or not by letting the user know of that extra parameter could lead to confusion. That is, the docs would need to explain how it is prioritising each of them in the conflicting cases like `LinearEqualityToPenalty(penalty=5, should_define_penalty=True)`, which is actually not considered in my PR as keeping `should_define_penalty` hidden avoids that situation.

Instead, explaining in the docs that if no `penalty` is given, a new value is calculated on every new conversion (which is done internally with `should_define_penalty` or whatever that variable is called) I think makes it more user-friendly. In the end, to the user what only matters is:

- if no `penalty` is given, the converter should calculate one
- if the `penalty` is given, it is the one that should be always used

But he doesn't care about the variables used by the converter to update the value, decide when to calculate, etc.

qiskit_optimization/converters/linear_equality_to_penalty.py
a-matsuo: Ah, I got it. The line is actually calling `@penalty.setter`.

qiskit_optimization/converters/linear_equality_to_penalty.py
a-matsuo: I see. `penalty` is set on line 125. I think it works better for the case that value is not updated as you mentioned.

a-matsuo:
> - if no penalty is given, the converter should calculate one
> - if the penalty is given, it is the one that should be always used
I agree with these specifications. I think the behavior is reasonable and it's ok as long as the behavior is written in the documentation.

peguerosdc:
> Wait a moment. After the first conversion, `self._penalty` is updated, so `should_define_penalty` is ignored in the second conversion?

Nope, because the condition is not `if self._penalty is None` anymore but `if self._should_define_penalty`, so as long as `should_define_penalty` is not updated, the penalty will be re-calculated on every new conversion.

a-matsuo (approved):

a-matsuo:
LGTM 🙂

woodsp-ibm:
Can unit test(s) be added to assure this function and that it continues to work as expected?

peguerosdc:
I have added one unit test to test the behaviour. It could be decomposed in smaller unit tests, but let me know what you think.

woodsp-ibm:
> I have added one unit test to test the behaviour.

Looks good thanks, since its testing the same object on a re-use basis it seems fine.

woodsp-ibm (approved):

PR #45, opened 2021-03-12
Interface for Gurobi through the gurobipy API

RichardOberdieck:
### Summary
Currently, qiskit only works with docplex as the LP/QP/MIP optimization engine. This pull request adds the ability to use Gurobi as well.

### Details and comments
Gurobi is a commercial optimization solver (like CPLEX). However, at the moment it is only possible to use cplex if one uses qiskit. As I see that qiskit will grow in the future, I believe that it adds value to enable the use of other optimization engines as well, specifically Gurobi in this case.

I was quite careful to preserve the style of the original codebase, as I mostly copied the `docplex` methods and replaced the API calls with the corresponding `gurobipy` calls. All the tests pass, and I added some where I found them to be relevant.

The only question I was unable to answer comes in `test_quadratic_program.py` in line 830-831 of my PR. Specifically, the LP formats between Gurobi and cplex differ, and therefore that test does not carry over from the corresponding test for `docplex`. Therefore I decided to comment it out as fixing it would require adding additional logic to the `__repr__` method of `QuadraticProgram`. I believe this is not very clean and therefore commented it out. If you have a better idea, then please do share.

Also, if you have any other comments/questions/remarks, please share.


stefan-woerner:
Hi @RichardOberdieck, thanks for your contribution! We'll have a look and get back to you soon. Small first comment, `max_cut.py` has been recently refactored and is now build around the `QuadraticProgram`, so you can remove that from your PR.

RichardOberdieck:
> Hi @RichardOberdieck, thanks for your contribution! We'll have a look and get back to you soon. Small first comment, `max_cut.py` has been recently refactored and is now build around the `QuadraticProgram`, so you can remove that from your PR.

Ok, sure. Shall I keep any of it or shall I just delete `max_cut.py` completely?

stefan-woerner:
@RichardOberdieck yes, the whole `ising` folder under `applications` has been deleted (i.e. moved and refactored), so you can delete `applications/ising/max_cut.py` or fetch from master, which should do it automatically (possibly via resolving a merge conflict). Thanks!

RichardOberdieck:
When I ran lint on my machine everything worked well. Are these failing checks a consequence of the merge conflicts? If so, shall I have a look at those?

manoelmarques:
Gurobipy requires an installation from a private Pypi site. Until it goes to the public Pypi it cannot be included in setup.py because it will fail any install that doesn't have guropbipy installed. I added an install of gurobipy to the CI in order to test it.
Also, the 'disp' property doesn't suppress all gurobipy output, first because the Model creation already print some stuff and LogToConsole should be set to 0 also: https://support.gurobi.com/hc/en-us/articles/360044784552-How-do-I-suppress-all-console-output-from-Gurobi-

manoelmarques:
Gurobipy is optional install so I also changed QuadraticProgram to make sure it works without it. However I am not sure this is the best place to have 'to_gurobipy' and 'from_gurobipy' methods. I had to change the typehints in those modules to Any since I can't add gurobipy.Model as typehin as it would fail if it is not installed

RichardOberdieck:
> Gurobipy requires an installation from a private Pypi site. Until it goes to the public Pypi it cannot be included in setup.py because it will fail any install that doesn't have guropbipy installed. I added an install of gurobipy to the CI in order to test it.

That's fair. Gurobi is currently in the process of getting to the public PyPI server, but there are some bureaucratic steps in the way. I'll let you know here when `gurobipy` is available from the public PyPI server.

> Also, the 'disp' property doesn't suppress all gurobipy output, first because the Model creation already prints some license stuff and LogToConsole should be set to 0 also: https://support.gurobi.com/hc/en-us/articles/360044784552-How-do-I-suppress-all-console-output-from-Gurobi-

The key here is actually that it is set during the environment creation, rather than on the model level. Will you fix this in the PR or shall I?

t-imamichi:
What is the license of `gurobipy`?

RichardOberdieck:
> What is the license of `gurobipy`?

When installing through pip, a non-production license limited to 2000 variables and linear constraints is issued. Otherwise, the user has to get a Gurobi license on www.gurobi.com. For academics, this is free. For commercial users, they will need to purchase a license from Gurobi.

t-imamichi:
Could you apply black by `make black` as explained https://github.com/Qiskit/qiskit-optimization/blob/main/CONTRIBUTING.md#pull-request-checklist
We recently apply black to the entire code base. See #117 for details.

qiskit_optimization/problems/quadratic_program.py
```diff
@@ -1056,6 +1301,7 @@ def export_as_lp_string(self) -> str:
         Returns:
             A string representing the quadratic program.
         """
+
```
t-imamichi, line 1304:
```suggestion
```

requirements.txt
```diff
@@ -6,3 +6,4 @@ docplex==2.15.194; sys_platform == 'darwin'
 setuptools>=40.1.0
 networkx>=2.2
 dataclasses; python_version < '3.7'
+gurobipy
```
t-imamichi, line 9: Gurobipy should be an extra library. So, this line needs to move `setup.py` similar to `cplex`.
https://github.com/Qiskit/qiskit-optimization/blob/f3a4df4e3fe75331b9475755462e5fae5e3254b8/setup.py#L64

t-imamichi:
We are discussing generalization of `QP.from_XXX` and `QP.to_XXX` in #116 #122. The discussion of the design is not finalized yet. We may separate your implementation as one of translators in the future.

requirements.txt
manoelmarques: Done

t-imamichi (approved):
LGTM

t-imamichi:
Could you write a release note following [the guide](https://github.com/Qiskit/qiskit-optimization/blob/main/CONTRIBUTING.md#adding-a-new-release-note)?

RichardOberdieck:
@t-imamichi, @manoelmarques : let me know if I can contribute to the release note.

t-imamichi:
Thanks. Could you check out the following guide and write a reno?
https://github.com/Qiskit/qiskit-optimization/blob/main/CONTRIBUTING.md#adding-a-new-release-note

t-imamichi:
@RichardOberdieck If you cannot write a reno, we can write one on be half of you and merge this PR.

RichardOberdieck:
@t-imamichi : I'd appreciate that. Due to the refactor of `quadratic_program.py` etc. I am not quite sure I could say anything except "QISKIT supports gurobipy now".

t-imamichi:
Thanks. I will do that.

t-imamichi (dismissed):

stefan-woerner (approved):
lgtm

t-imamichi (approved):

PR #50, opened 2021-03-18
Fixed a bug in vehicle_routing, and added test cases for vehicle_routing and for clique. 

tamiya-onodera:
### Summary

Fixed a bug in vehicle_routing,  while added a test case for the fix, another test case for vehicle_routing, and a test case for clique.

### Details and comments

Fixed a bug in the 'interpret' function of vehicle_routing, which caused the right solution not to be produced when the node index for the depot is not the default value. 

Added a test case for vehicle_routing in which the depot is not the default value, and another test case in which the number_of vehicles is not the default value. 

Also added a test case for clique in which the clique size is specified.


a-matsuo:
Thank you for fixing the bug and adding unittests. LGTM!

woodsp-ibm (approved):

PR #51, opened 2021-03-19
Fix node colors of application classes

t-imamichi:
### Summary

Fix the node colors of application classes.
Move Gset to maxcut.
Adjusted requirements.txt and setup.py to remove unused libraries and add necessary libraries. 

### Details and comments

Nodes of graph of networkx are not always sorted. We need to set colors in the same order of nodes.

Gset is a benchmark dataset specific to maxcut. So, I move it to Maxcut class.
https://medium.com/toshiba-sbm/benchmarking-the-max-cut-problem-on-the-simulated-bifurcation-machine-e26e1127c0b0

```python
from qiskit.algorithms import QAOA, NumPyMinimumEigensolver
from qiskit_optimization.applications.vertex_cover import VertexCover
import networkx as nx

seed = 123
graph = nx.random_regular_graph(d=3, n=6, seed=seed)
pos = nx.spring_layout(graph, seed=seed)
prob = VertexCover(graph)
qp = prob.to_quadratic_program()
meo = MinimumEigenOptimizer(min_eigen_solver=NumPyMinimumEigensolver())
result = meo.solve(qp)
print(result)
prob.draw(result, pos=pos)
```

master
image

this pr
image


mtreinish (commented):
I think we should probably add networkx to the requirements list in the setup.py, it probably should have been added in #31 when the dependency was added but we should fix that before release. We're probably not catching this in tests because it's in the requirements-dev.txt, but when users go to use these classes  they'll get an import error because networkx is not installed.

As an aside could you also open issues on retworkx for the functions which are missing there that prevent you from using it? The only one I saw from skimming through the code was `complement() ` which should be simple enough to add. (these new graph optimization algorithm classes are the only place in Qiskit using networkx)

t-imamichi:
Thank you @mtreinish. I fix setup.py to include networkx.

We use networkx because of `draw`. Do you plan to support it in retworkx?

mtreinish:
> Thank you @mtreinish. I fix setup.py to include networkx.
> 
> We use networkx because of `draw`. Do you plan to support it in retworkx?

Sure we can look at adding it to retworkx. I had previously avoided adding a matplotlib drawer because it would be an optional dependency and a python module instead of rust function, but also because the [networkx documentation](https://networkx.org/documentation/stable//reference/drawing.html) recommends using a graphviz based drawer. But if you have a use case for it, lets open an issue to track adding a `draw` function built using matplotlib and we can come up with an implementation. Making sure the library is usable and fits our use cases is more important.

FWIW, right now retworkx has drawing support using graphviz and the `to_dot` method: https://retworkx.readthedocs.io/en/stable/stubs/retworkx.PyGraph.html#retworkx.PyGraph.to_dot has an example on how to generate an image from a retworkx graph using `to_dot` and [pydot](https://pypi.org/project/pydot/). You could mirror the existing `draw` usage with it.

a-matsuo:
Thank you for fixing the bug. Could you please give me reproducible test cases? Because I thought those nodes are sorted and I have used `graph.nodes` or `graph.edge` a lot for iterations of nodes or edges in a graph when I make a corresponding quadratic program. I might need to change those `graph.nodes` or `graph.edge` as well.

t-imamichi:
I attached the code in the descrition.

t-imamichi:
@a-matsuo @manoelmarques I added matplotlib to requirements.txt and setup.py because networkx.draw depends on matplotlib.
https://networkx.org/documentation/stable//reference/drawing.html

a-matsuo (approved):

requirements.txt
```diff
@@ -1,9 +1,9 @@
 qiskit-terra>=0.17.0
 scipy>=1.4
 numpy>=1.17
-psutil>=5
 docplex; sys_platform != 'darwin'
 docplex==2.15.194; sys_platform == 'darwin'
 setuptools>=40.1.0
-retworkx>=0.7.0
+networkx>=2.2
+matplotlib
```
woodsp-ibm, line 8: I did not see any code that was added here that depends on `matplotlib` did I miss something? I get that we would prefer networkx at this application level since users know that better

t-imamichi: `networkx.draw` depends on matplotlib.
https://networkx.org/documentation/stable//reference/drawing.html

This line also depends on matplotlib
https://github.com/Qiskit/qiskit-optimization/blob/d0ffed7418ac01a14b96bdf3ae6ff4e211193d9e/qiskit_optimization/applications/vehicle_routing.py#L19

woodsp-ibm: matplotlib is a dependent of a dependent. If you look networkx has it as an extra_requires meaning if you do not want to use draw you not need to install it. So if a user at our level does not want to use draw we should not force it to be installed in my view if its optional in the dependent. I would not be adverse to having some extra_requires at our level - and state in the docs that to draw you need this.

woodsp-ibm: As to the vehicle_routing it seems to need it solely for some color definition `edge_cmap=plt.cm.plasma` which presumably we should be able to do without requiring matplotlib to define some colors for this purpose

t-imamichi: I think draw is one of the main features of the application classes. So, we need more discussion to make matplotlib extra_requires. I will make another issue for that purpose. But, I would like to merge this pr as is because matplotlib is currently required by draw.

woodsp-ibm (approved):

PR #69, opened 2021-04-02
Fix Doc Theme Errors

manoelmarques:


woodsp-ibm (approved):
Decorator addition is due to an incompatibility with networkx with the latest release of decorator. Hopefully change can be reverted once that is corrected

PR #71, opened 2021-04-03
Remove decorator from requirements, added during its buggy period

manoelmarques:
### Details and comments

Decorator had a backward compatibility problem that was affecting networkx for 2 days but the intermediate buggy versions were yanked and a fix was released.
In the meantime, I had added a pin to decorator in requirements.txt in order to publish our docs which is unnecessary now. I am reverting to as it used to be.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,5 @@ numpy>=1.17
 docplex; sys_platform != 'darwin'
 docplex==2.15.194; sys_platform == 'darwin'
 setuptools>=40.1.0
-networkx>=2.2
-decorator<5.0
+networkx>=2.5.1
 dataclasses; python_version < '3.7'
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,5 @@ numpy>=1.17
 docplex; sys_platform != 'darwin'
 docplex==2.15.194; sys_platform == 'darwin'
 setuptools>=40.1.0
-networkx>=2.5.1
+networkx>=2.2
 dataclasses; python_version < '3.7'


dschult:
Decorator v5.0.5 fixes the problem. So you can now stay with networkx2.5 using decorator5.0.5 or which to networkx2.5.1 using decorator4

manoelmarques:
The issue is that we don't use decorator in the code. Then we have to add a dependency to decorator pinning it to > 5.0.5 just to fix a problem that is inherent to networkx and has nothing to do with us. I would rather pin networkx>2.5.1  which is the only thing we use and let them deal with decorator past/future changes than to add a dependency we don't use. The reason decorator is in this list is because I temporarily included it just to finish this latest release.

dschult:
Right...  and that's a reasonable approach. I just wanted to let you know that you could leave your code the way it was and the NetworkX issue caused downstream by decorator should now fix itself since 5.0.5 will be chosen over 5.0.0-5.0.4 by pip or Fonda or whatever...   either way should work. The decorator issue was unfortunate, unanticipated and now solved on their end. It shouldn't cause any further problems.

manoelmarques:
Yes, I see that decorator yanked 5 versions in a roll. So it seems only the people who already installed decorator during the 2 days buggy period would be affected if I just remove decorator from requirements.txt and leave networkx>=2.2 as it used to be. We have to decide here if we leave networkx>=2.2 or change to networkx>=2.5.1.

woodsp-ibm (approved):

PR #93, opened 2021-04-21
Updated example in README file

pawmasz:
### Summary
Updated README file to show example with correct arguments in two functions.


woodsp-ibm (approved):
Thanks for the fix. We have a test file that supposed to ensure this keeps working https://github.com/Qiskit/qiskit-optimization/blob/main/test/test_readme_sample.py where the body of the test is the readme sample here. Seems it got updated without the readme being changed - but your fix corrects things and now they look the same.

PR #96, opened 2021-04-22
Fix unit test of GroverOptimizer

a-matsuo:
### Summary
Fixes #88 

### Details and comments
As mentioned in #88, somehow, we get the different results even with the same seed. This issue didn't occur before. 
I used a different example for the unit test and I get the same result in that case.

PR #97, opened 2021-04-22
Fix bit ordering and probabilities of samples in optimization_algorithm.py

### LINKED ISSUE #92: MinimumEigenOptimizer returns a wrong solution
### Information

- **Python version**: 3.8.9
- **Operating system**: macOS 11.2.3

`MinimumEigenOptimizer` with qasm_simulator returns a wrong solution.

```python
from docplex.mp.model import Model
mdl = Model("docplex model")
x = mdl.binary_var('x')
y = mdl.binary_var('y')
mdl.minimize(x-2*y)
print(mdl.export_as_lp_string())
result = mdl.solve()
print("result of CPLEX solver")
print(result)

from qiskit.providers.aer import QasmSimulator
from qiskit.algorithms.optimizers import SPSA
from qiskit.circuit.library import TwoLocal
from qiskit.utils import QuantumInstance
from qiskit.algorithms import VQE
seed = 1234
qp = QuadraticProgram()
qp.from_docplex(mdl)
backend = QasmSimulator()
optimizer = SPSA(maxiter=100)
ry = TwoLocal(2, 'ry', 'cz', reps=3, entanglement='full')
quantum_instance = QuantumInstance(backend=backend, seed_simulator=seed, seed_transpiler=seed)
vqe_mes = VQE(ry, optimizer=optimizer, quantum_instance=quantum_instance)
vqe = MinimumEigenOptimizer(vqe_mes)
result = vqe.solve(qp)
print("result of VQE solver")
print(result)
print(result.raw_samples)
```

output:
```
\ ENCODING=ISO-8859-1
\Problem name: docplex model

Minimize
 obj: x - 2 y
Subject To

Bounds
 0 <= x <= 1
 0 <= y <= 1

Binaries
 x y
End

result of CPLEX solver
solution for: docplex model
objective: -2
y=1

result of VQE solver
optimal function value: 1.0
optimal value: [1. 0.]
status: SUCCESS
```

The solution of VQE should be `[0. 1.]`.

### Suggested solutions

We may need to reverse the bitstrings as `x = np.fromiter(list(bitstr[::-1]), dtype=int)`. But, it may affect other algorithms and unit tests.
https://github.com/Qiskit/qiskit-optimization/blob/54a6d0750492132aa019975e3a8648c507b7553c/qiskit_optimization/algorithms/optimization_algorithm.py#L511

a-matsuo:
### Summary

- Fix bit ordering issue and add unit tests.
- Fix probabilities of samples.

Fixes #92

### Details and comments
This PR should be merged after #96 

#### Bit ordering
We need to reverse bit ordering from qiskit bit ordering to conventional bit ordering for the actual outputs.

#### Probabilities of samples
This PR fixes the following lines that  calculate the probabilities of solutions. 
When `eigenvector` is a dict, we need to square the values since the values are normalized.
https://github.com/Qiskit/qiskit-optimization/blob/724158ba42f4b75758edc1de42f6f665660ec27e/qiskit_optimization/algorithms/optimization_algorithm.py#L554-L557

I will explain it in details for the record. 
Originally, when we were using Qiskit Aqua, this `eigenvector` was supposed  to be the resultant `counts` of a qasm simulator. Thus, the above code was working correctly.
https://github.com/Qiskit/qiskit-aqua/blob/aa8bc0c8011405d382fc05d6ab576dd926bdd2a7/qiskit/aqua/algorithms/minimum_eigen_solvers/vqe.py#L578

However, after the code was moved into Qiskit terra, there was some bug and it was fixed in  https://github.com/Qiskit/qiskit-terra/pull/5496 .
As a result, this `eigenvector` became the square root of the normalized counts, i.e. from `{bitstr: counts}` to `{bitstr: np.sqrt(counts / shots)}` .
https://github.com/Qiskit/qiskit-terra/blob/3c979ebdacecbbd213757a8149bddc309bfa58c0/qiskit/algorithms/minimum_eigen_solvers/vqe.py#L520

If we change the value of this `eigenvector` in Qiskit terra in the future, we need to change the above code in `MinimumEigenSolvers` as well.


test/algorithms/test_min_eigen_optimizer.py
```diff
@@ -245,6 +252,26 @@ def test_samples(self):
...
+        optimizer = SPSA(maxiter=100)
+        ry = TwoLocal(2, 'ry', 'cz', reps=3, entanglement='full')
+        quantum_instance = QuantumInstance(backend=backend,
+                                           seed_simulator=seed, seed_transpiler=seed)
+        vqe_mes = VQE(ry, optimizer=optimizer, quantum_instance=quantum_instance)
+        vqe = MinimumEigenOptimizer(vqe_mes)
+        result = vqe.solve(qp)
+        self.assertEqual(result.fval, -2)
+        self.assertListEqual(result.x.tolist(), [0, 1])
+
```
t-imamichi, line 274: Could you add some tests with samples? The optimal solution should have the highest probability.

a-matsuo: Sure. I added a test with samples as well.

t-imamichi:
We don't need to wait for #96 because CI builds Aer on Ubuntu.

test/algorithms/test_min_eigen_optimizer.py
t-imamichi: It would be better to follow `test_samples`, for example, comparison of x should be `np.testing.assert_array_almost_equal`.

t-imamichi: Why `result.raw_samples[0].x[0:2]` instead of `result.raw_samples[0].x`?

a-matsuo: Sure. I changed to use `np.testing.assert_array_almost_equal` and fixed the slice as you mentioned (I didn't need it in this test case)

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -236,6 +235,21 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
                     logger.info('Current Optimum Value: %s', optimum_value)
                     improvement_found = True
                     threshold = optimum_value
+
+                    # trace out work qubits and store samples
```
t-imamichi, line 239: Is this change also related to bit ordering?

a-matsuo: This comes from [#96](
https://github.com/Qiskit/qiskit-optimization/pull/96/files) because I expected this PR to be merged after #96 since 
the unittest of `GroverOptimizer` was failing and CI was not working. I think we'd better merge #96 before this PR since `GroverOptimizer` also relies on `generate_solution`.

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -313,7 +313,7 @@ def _get_probs(self, qc: QuantumCircuit) -> Dict[str, float]:
             state = result.get_counts(qc)
             shots = self.quantum_instance.run_config.shots
             hist = {key[::-1]: val / shots for key, val in state.items() if val > 0}
-            self._circuit_results = {b[::-1]: np.sqrt(v / shots) for (b, v) in state.items()}
+            self._circuit_results = {b: np.sqrt(v / shots) for (b, v) in state.items()}
```
t-imamichi, line 316:
```suggestion
            self._circuit_results = {b[::-1]: np.sqrt(v / shots) for (b, v) in state.items()}
```

a-matsuo: I accidentally removed this code `self._circuit_results = {i[-n_key:]: v for i on line 247 when I removed the change of #96. That was causing an error. Please revert this change. Line 316 needs to be `self._circuit_results = {b: np.sqrt(v / shots) for (b, v) in state.items()}`

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -313,7 +313,7 @@ def _get_probs(self, qc: QuantumCircuit) -> Dict[str, float]:
             state = result.get_counts(qc)
             shots = self.quantum_instance.run_config.shots
             hist = {key[::-1]: val / shots for key, val in state.items() if val > 0}
-            self._circuit_results = {b: np.sqrt(v / shots) for (b, v) in state.items()}
+            self._circuit_results = {b[::-1]: np.sqrt(v / shots) for (b, v) in state.items()}
```
t-imamichi:
```suggestion
            self._circuit_results = {b: np.sqrt(v / shots) for (b, v) in state.items()}
```

qiskit_optimization/algorithms/grover_optimizer.py
t-imamichi: I reverted it according to your suggestion.

t-imamichi: After I reverted it, unit tests of Grover fail. Are you sure `self._circuit_results = {b: np.sqrt(v / shots) for (b, v) in state.items()}` is correct?

test/algorithms/test_min_eigen_optimizer.py
```diff
@@ -245,6 +252,28 @@ def test_samples(self):
         self.assertAlmostEqual(result.fval, result.samples[0].fval)
         self.assertEqual(result.status, result.samples[0].status)
 
+    def test_bit_ordering(self):
```
t-imamichi, line 255: Could you add test of bit ordering for Grover and statevector too?

qiskit_optimization/algorithms/grover_optimizer.py
a-matsuo: Yaya, we also need to add `self._circuit_results = {i[-n_key:]: v for i` as I commented in the above though.

test/algorithms/test_min_eigen_optimizer.py
a-matsuo: Sure. I added unittests for GroverOptimizer and statevector as well.

a-matsuo:
Somehow, there is a large gap between the results of `GroverOptimizer` with my local environment and those with Python 3.6 in CI. (The correct answer is [1, 0, 0, 0, 0] so it should have larger probability than others.). Is the change #96 already included in CI?

My local environment with Python 3.8.6
```
SolutionSample(x=array([1, 0, 0, 0, 0]), fval=1.0, probability=0.09709528839529034, status=<OptimizationResultStatus.SUCCESS: 0>)
SolutionSample(x=array([1, 0, 1, 0, 1]), fval=9.0, probability=0.03607710007007227, status=<OptimizationResultStatus.SUCCESS: 0>)
SolutionSample(x=array([0, 0, 1, 0, 1]), fval=26.0, probability=0.034583601441991185, status=<OptimizationResultStatus.SUCCESS: 0>)
```

Python 3.6 in CI
```
SolutionSample(x=array([1, 0, 0, 0, 1]), fval=25.0, probability=0.03725333902474391, status=<OptimizationResultStatus.SUCCESS: 0>)
SolutionSample(x=array([1, 1, 1, 0, 0]), fval=59.0, probability=0.036416078422458094, status=<OptimizationResultStatus.SUCCESS: 0>)
SolutionSample(x=array([0, 1, 0, 0, 1]), fval=8.0, probability=0.035559109477723554, status=<OptimizationResultStatus.SUCCESS: 0>)
```

woodsp-ibm:
> Is the change #96 already included in CI?

That PR has already been merged and hence your PR, which is up todate with main, is included in CI.

woodsp-ibm:
Ok, I see you suppressed the test, I will note that the latest PR from Manoel also sets a flag that would allow it to run the set of tests even if it failed for one python version so we could have seen if it was just 3.6 in CI or all the versions. We were going to merge that and then try re-running this to see whether it passed on 3.8 like it did locally for you, or fails in the other Python version too

a-matsuo:
That's nice! I noticed  it's happening on other versions as well. hmm...

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -227,6 +226,8 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
                 int_v = self._bin_to_int(v, n_value) + threshold
                 logger.info('Outcome: %s', outcome)
                 logger.info('Value Q(x): %s', int_v)
+                print("rt", rotation_count)
```
t-imamichi, line 229: Remove `print`. If necessary use `logger`.

a-matsuo:
Whoa! I changed to use BasicAer instead of Aer in the unittest of `GroverOptimizer`. Then, I can see the probability of the correct solution ([1, 0, 0, 0, 0]) is beautifully increased. Before changing to BasicAer, something was strange. Even though `GroverOptimizer` increased the number of the iterations of `Grover` (so the probability of the correct solution should increase), the probability distribution did not change at all, literally. It stayed at a uniform distribution. 

Python 3.6 in CI with BasicAer
```
SolutionSample(x=array([1, 0, 0, 0, 0]), fval=1.0, probability=0.6123046875, status=<OptimizationResultStatus.SUCCESS: 0>)
237
SolutionSample(x=array([0, 0, 1, 0, 0]), fval=2.0, probability=0.015625, status=<OptimizationResultStatus.SUCCESS: 0>)
238
SolutionSample(x=array([0, 1, 0, 1, 0]), fval=2.0, probability=0.009765625, status=<OptimizationResultStatus.SUCCESS: 0>)
```

qiskit_optimization/algorithms/grover_optimizer.py
a-matsuo: Don't worry. I will remove all the "print"s in the end. I just wanted to see what's happening in the CI.

woodsp-ibm:
I don't know - main branch here will be Aer built from its master branch. Its possible that some change since release affected things. Are you able to install Aer from master locally an try - if it fails I think we should raise an issue on Aer

a-matsuo:
Finally, I figured out what was going on. Algorithms and Qiskit Aer were ok. I was really unlucky (or lucky) and I chose the seed  which fails in CI (Failure seed exist with a probability of 0.2%). I increase `num_iterations`, so  even Aer produces a different result with the same seed, it is very unlikely to fail in the future (0.006%). Let me add more tests and finalize the PR.

t-imamichi:
Could you add `test_bit_ordering` to `test_grover_optimizer.py`?

a-matsuo:
Ok, done! Could you guys review the PR plz?

test/algorithms/test_min_eigen_optimizer.py
```diff
@@ -156,91 +180,136 @@ def test_converter_list(self):
...
         self.assertAlmostEqual(result.raw_samples[0].probability, 1.0)
-        self.assertEqual(result.raw_samples[0].status, SUCCESS)
+        self.assertEqual(result.raw_samples[0].status, success)
 
-        qaoa = QAOA(quantum_instance=quantum_instance)
+    @data('sv', 'qasm')
+    def test_samples_qaoa(self, simulator):
+        """Test samples for QAOA"""
+        # test minimize
+        print("qaoa")
```
t-imamichi, line 223:
```suggestion
```

test/algorithms/test_min_eigen_optimizer.py
```diff
@@ -156,91 +180,136 @@ def test_converter_list(self):
...
+        self.assertEqual(result.samples[0].status, success)
+        np.testing.assert_array_almost_equal(result.raw_samples[0].x, [0, 0, 1, 0])
+        self.assertAlmostEqual(result.raw_samples[0].fval, opt_sol)
+        self.assertEqual(result.raw_samples[0].status, success)
+
+    @data('sv', 'qasm')
+    def test_samples_vqe(self, simulator):
+        """Test bit ordering"""
+        # test minimize
+        print(simulator)
```
t-imamichi, line 276:
```suggestion
```

test/algorithms/test_min_eigen_optimizer.py
```diff
@@ -156,91 +180,136 @@ def test_converter_list(self):
...
+        x = mdl.binary_var('x')
+        y = mdl.binary_var('y')
+        mdl.minimize(x-2*y)
+        op = QuadraticProgram()
+        op.from_docplex(mdl)
+        opt_sol = -2
+        success = OptimizationResultStatus.SUCCESS
+        optimizer = SPSA(maxiter=100)
+        ry_ansatz = TwoLocal(5, 'ry', 'cz', reps=3, entanglement='full')
+        # Test for a qasm simulator.
```
t-imamichi, line 289:
```suggestion
```

t-imamichi:
Did you test `min x - 2 * y` with `GroverOptimizer` too?

a-matsuo:
@t-imamichi Thank you for the comments. I fixed them. Could you please check files again?

test/algorithms/test_min_eigen_optimizer.py
```diff
@@ -168,98 +198,145 @@ def test_converter_list(self):
...
+        self.assertEqual(result.status, success)
+        np.testing.assert_array_almost_equal(result.samples[0].x, [0, 1])
+        self.assertAlmostEqual(result.samples[0].fval, opt_sol)
+        self.assertEqual(result.samples[0].status, success)
+        np.testing.assert_array_almost_equal(result.raw_samples[0].x, [0, 0, 1, 0])
+        self.assertAlmostEqual(result.raw_samples[0].fval, opt_sol)
+        self.assertEqual(result.raw_samples[0].status, success)
+
+    @data("sv", "qasm")
+    def test_samples_vqe(self, simulator):
```
t-imamichi, line 296: `test_bit_ordering` would be good to be consistent with the comment below, "Test bit ordering"

t-imamichi:
Could you apply `black -l 100` to align with #118?

test/algorithms/test_min_eigen_optimizer.py
a-matsuo: Actually, I changed the comment of `test_samples_vqe` to `"""Test samples for VQE"""`. Since the names of other test cases are `test_samples_qaoa` and `test_samples_numpy_eigen_solver`,  `test_samples_vqe` looks more consistent than `test_bit_ordering`.

qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -551,10 +552,9 @@ def generate_solution(bitstr, qubo, probability):
 
         solutions = []
         if isinstance(eigenvector, dict):
-            all_counts = sum(eigenvector.values())
+            eigenvector = {bitstr: val ** 2 for (bitstr, val) in eigenvector.items()}
```
t-imamichi, line 555: Instead of overwriting `eigenvector`, saving as `probabilities` seems more consistent with the case `isinstance(eigenvector, np.ndarray)`. What do you think?
```
            probabilities = {bitstr: val ** 2 for (bitstr, val) in eigenvector.items()}
```

t-imamichi:
Could you add bit ordering test of QAOA just in case?

qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -552,9 +552,9 @@ def generate_solution(bitstr, qubo, probability):
 
         solutions = []
         if isinstance(eigenvector, dict):
-            eigenvector = {bitstr: val ** 2 for (bitstr, val) in eigenvector.items()}
+            probabilities = {bitstr: val ** 2 for (bitstr, val) in probabilities.items()}
```
t-imamichi, line 555:
```suggestion
            probabilities = {bitstr: val ** 2 for (bitstr, val) in eigenvector.items()}
```

t-imamichi (approved):
LGTM. Thanks!

woodsp-ibm:
Does this in anyway depend on Qiskit/qiskit-terra#6365 ? Since all the unit tests pass here I am guessing not, but that is another reversal of order right?

a-matsuo:
>Does this in anyway depend on Qiskit/qiskit-terra#6365 ? Since all the unit tests pass here I am guessing not, but that is another reversal of order right?

That's another reversal of order, but actually this does not depend on Qiskit/qiskit-terra#6365. It's simply a bug caused by inconsistency. Currently, `oracle_evaluation` in `PhaseOracle` supports the conventional bit order, so I'm planning to change it to support the qiskit bit order in Qiskit/qiskit-terra#6365.

t-imamichi (dismissed):

t-imamichi (approved):

PR #110, opened 2021-04-28
Fix multistart optimizer.py number of trials

### LINKED ISSUE #106: Check whether the number of trials is positive or not in `MultiStartOptimizer`

`MultiStartOptimizer ` has a parameter `_trials` and it should be positive (> 0). Otherwise, it fails in the main loop.
We need to check it when `_trials` is set as follows. 

https://github.com/Qiskit/qiskit-optimization/blob/51628c27f70c042adf93bedbfee16eed09a53e1b/qiskit_optimization/algorithms/multistart_optimizer.py#L55
https://github.com/Qiskit/qiskit-optimization/blob/51628c27f70c042adf93bedbfee16eed09a53e1b/qiskit_optimization/algorithms/multistart_optimizer.py#L115

Discussed in https://github.com/Qiskit/qiskit-optimization/issues/41#issuecomment-825715102

alimustafashah:
Fixes #106

### DIFF
--- a/qiskit_optimization/algorithms/multistart_optimizer.py
+++ b/qiskit_optimization/algorithms/multistart_optimizer.py
@@ -53,7 +53,10 @@ class MultiStartOptimizer(OptimizationAlgorithm, ABC):
         """
         super().__init__()
         
-        self._trials = trials if trials > 0 else 1
+        if trials <= 0:
+            raise ValueError('Number of trials should be positive')
+        
+        self._trials = trials
         self._clip = clip
 
     def multi_start_solve(self, minimize: Callable[[np.ndarray], Tuple[np.ndarray, Any]],
@@ -113,8 +116,10 @@ class MultiStartOptimizer(OptimizationAlgorithm, ABC):
         Args:
             trials: The number of trials to set.
         """
-        if trials > 0:
-            self._trials = trials
+        if trials <= 0:
+            raise ValueError('Number of trials should be positive')
+            
+        self._trials = trials
 
     @property
     def clip(self) -> float:
--- a/qiskit_optimization/algorithms/multistart_optimizer.py
+++ b/qiskit_optimization/algorithms/multistart_optimizer.py
@@ -52,7 +52,8 @@ class MultiStartOptimizer(OptimizationAlgorithm, ABC):
                 with the ``-clip`` or ``clip`` values correspondingly for the initial guesses.
         """
         super().__init__()
-        self._trials = trials
+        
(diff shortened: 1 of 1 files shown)


qiskit_optimization/algorithms/multistart_optimizer.py
```diff
@@ -52,7 +52,8 @@ def __init__(self, trials: int = 1, clip: float = 100.) -> None:
                 with the ``-clip`` or ``clip`` values correspondingly for the initial guesses.
         """
         super().__init__()
-        self._trials = trials
+        
+        self._trials = trials if trials > 0 else 1
```
woodsp-ibm, line 56: If trials <= 0 is really an error in regards of not being a valid value should we rather raise a ValueError, rather than 'hiding' the fact that this is being configured incorrectly i.e. by forcing it to 1 in the constructor or ignoring it as in the setter.

alimustafashah: We can do it both ways. Tell me if you want to raise error. Can do that too!

woodsp-ibm: @adekusar-drl Any thought on the above?

alimustafashah: Since the default value of trials is 1 in the constructor so, when the user gives non-positive value if we assign 1 to self._trials would not be something completely unexpected. On the other hand, your point of 'hiding' the fact is also valid. Just tell me which approach you want to go with and I will do it!

qiskit_optimization/algorithms/multistart_optimizer.py
```diff
@@ -112,7 +113,8 @@ def trials(self, trials: int) -> None:
         Args:
             trials: The number of trials to set.
         """
-        self._trials = trials
+        if trials > 0:
```
t-imamichi, line 116: Could you raise a ValueError if `trials <= 0` as same as https://github.com/Qiskit/qiskit-optimization/pull/110/files#r622521214?

qiskit_optimization/algorithms/multistart_optimizer.py
adekusar-drl: I'd agree, raising an exception is better.

qiskit_optimization/algorithms/multistart_optimizer.py
alimustafashah: Yeah sure.

alimustafashah:
I am making a new pull request!

Issue #112, opened 2021-04-28
Adjust plot_histogram to other datasets


pawmasz:
### What is the expected enhancement?
`plot_histogram` function doesn't have an argument to change the y axis label for plots other than probabilities. It would also be good to adjust y axis values if they don't sum to 1, like optimized function values:

![image](https://user-images.githubusercontent.com/42217706/116481923-72329880-a884-11eb-9864-d6f27894f916.png)

Issue #81, opened 2021-04-14
Add visualization of solution samples


t-imamichi:
### What is the expected behavior?

Optimization results include not only the best solution but also the solution samples.
It would be nice to visualize the distribution of the solutions such as a histogram like [plot_histogram](https://qiskit.org/documentation/tutorials/circuits/2_plotting_data_in_qiskit.html#Plot-histogram).

![image](https://user-images.githubusercontent.com/31178928/114679569-3c02fe00-9d47-11eb-8000-c65961c1e546.png)

pawmasz:
@t-imamichi can I take a look at this?

t-imamichi:
Sure. Feel free to work on it.

MartinBeseda:
@pawmasz @t-imamichi Is this issue still being worked on? If it isn't I'd take a look at it - if it is, I'm sorry to bother your both.

Martin

pawmasz:
Go ahead 😉

t-imamichi:
I think the examples in #152 are enough for most users to visualize the samples.

PR #113, opened 2021-04-28
[WIP] Add get_probabilities method

pawmasz:
### Summary
Added `get_counts` method. This adds functionality requested in issue #81 

### Details and comments
`get_count` method for `OptimizationResult` class returns dictionary with samples, and function values. This allows to plot the counts using `plot_histogram` function:

image

Currently the label of y axis for `plot_histogram` can't be changed. I requested enchancment in issue #112


t-imamichi:
Thank you for making a PR. I have some comments.

First, #81 is originally thought to plot the distribution of the solutions, i.e., "x-axis: x value" v.s. "y-axis: probability". It would be also interesting to plot the distribution of the objective function values, i.e., "x-axis: fval" v.s. "y-axis: probability". Anyways, the y-axis should be always probabilities corresponding to `x` or `fval`.

Second, it would be good to add some utility functions directly plot figures of the distribution by calling `plot_histogram` not by generating an intermediate object.  It is because the returned object of `get_counts` has less information than `OptimizationResult` and not likely to be used anywhere else.  What do you think?

pawmasz:
Thanks for your input.

I definitely can do `x.value` vs. `probability`, and `fval` vs. `probability`. I'll work on it.

About the intermediate object: I followed the code of conduct in other `qiskit` modules. It is easier to import `plot_histogram` in your main code, create intermediate object that has the results, and then plot that.

If you want to explore the **direct** plot function, I can check possible solutions, but I lean towards the intermediate object.

t-imamichi:
OK. I think the intermediate object is fine. Although the extension to various types of plot is interesting, it might be good to focus on the basic feature as the first step. I have three requests.

1. renames the method as `get_probabilities` instead of `get_counts`.
2. returns a dict of {x-value: probability}. Then, it is compatible with `plot_histogram`. I'm not sure what is the best string format for "x-value". Could you try some ideas?
3. adds `status` to filter the solution. You currently picks up only succeeded solutions. But, we may want to plot solutions with different status or all statuses. For example, if arg `status=None`, we returns an object with all solutions of any status.

As a summary, I think the following signature is good to get the probabilities of solutions. `OptimizationResult.get_probabilities(self, status: Optional[OptimizationResultStatus]=None) -> Dict[str, float]`

pawmasz:
I added commits to reflect your requests. Current output of `get_probabilities` function:

image

t-imamichi:
Could you think of unit tests?

x-values can be not only binary values but also integer or continuous values.

qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -196,6 +196,26 @@ def get_correlations(self) -> np.ndarray:
...
+        """
+        sample_results = {}
+        if status is None:
+            for sample in self.samples:
+                sample_values = f"{''.join(map(str, map(int, sample.x)))}"
+                sample_results.update({sample_values: sample.probability})
+        else:
+            for sample in self.samples:
+                sample_values = f"{''.join(map(str, map(int, sample.x)))}"
+                if sample.status.name in status:
```
t-imamichi, line 214: You can simplify the code by checking the condition as follows.
```
if status is None or sample.status == status:
```
As for `status` arg, it should be `Optional[OptimizationResultStatus]`. We need to take advantage of enum.

qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -196,6 +196,26 @@ def get_correlations(self) -> np.ndarray:
...
+            A dictionary with sample names, and function values.
+        """
+        sample_results = {}
+        if status is None:
+            for sample in self.samples:
+                sample_values = f"{''.join(map(str, map(int, sample.x)))}"
+                sample_results.update({sample_values: sample.probability})
+        else:
+            for sample in self.samples:
+                sample_values = f"{''.join(map(str, map(int, sample.x)))}"
```
t-imamichi, line 213: `sample.x` is not always binary values. How do you distinguish `x=[1, 0, 1]`, `x=[10, 1]` and `x=[101]`?
I also recommend to use list comprehension instead of `map` two times.

pawmasz:
> Could you think of unit tests?
> 
> x-values can be not only binary values but also integer or continuous values.

Are x-values of different type than List when not binary? That would make it easy to distinguish.

qiskit_optimization/algorithms/optimization_algorithm.py
pawmasz: Thanks for the tip about condition.

How can we use `Optional[OptimizationResultStatus]` to choose the `status` that we're interested in, like in example below with current behavior.

image

t-imamichi:
This is the example that include both integer and continuous
```
from qiskit_optimization.algorithms import CobylaOptimizer
from qiskit_optimization.algorithms.admm_optimizer import ADMMOptimizer, ADMMParameters
from qiskit.algorithms.optimizers import COBYLA
from qiskit import Aer

qp = QuadraticProgram()
qp.integer_var(lowerbound=0, upperbound=3)
qp.continuous_var(lowerbound=0, upperbound=1.5)
qp.maximize(linear={0: 1, 1: 2})
print(qp)
solver = ADMMOptimizer(params=ADMMParameters(), continuous_optimizer=CobylaOptimizer())
result = solver.solve(qp)
for s in result.samples:
    print(s)
```

```
\ ENCODING=ISO-8859-1
\Problem name: CPLEX

Maximize
 obj: x0 + 2 x1
Subject To

Bounds
       x0 <= 3
       x1 <= 1.500000000000

Generals
 x0
End

SolutionSample(x=array([0. , 1.5]), fval=3.0, probability=1.0, status=<OptimizationResultStatus.SUCCESS: 0>)
```

t-imamichi:
As #152 is merged, we have a tutorial how to visualize solution samples. So, I want to close this PR. Do you have any comment, @pawmasz?

pawmasz:
> As #152 is merged, we have a tutorial how to visualize solution samples. So, I want to close this PR. Do you have any comment, @pawmasz?

Nothing from my side 😉

t-imamichi:
Thank you for your contribution. I then close this PR.

PR #114, opened 2021-04-29
Add maximize to minimize converter

### LINKED ISSUE #41: Maximize to minimize converter

Some algorithms convert maximization problems into minimization problems internally. By introducing `MaximizeToMinimize` converter, we can simplify the codes of the algorithms by adding `MaximizeToMinimize` to the default converter list.

https://github.com/Qiskit/qiskit-optimization/blob/57b59c64bf0ba72057465f2a6ff3560b26239ced/qiskit_optimization/algorithms/cobyla_optimizer.py#L115
https://github.com/Qiskit/qiskit-optimization/blob/57b59c64bf0ba72057465f2a6ff3560b26239ced/qiskit_optimization/algorithms/minimum_eigen_optimizer.py#L198
https://github.com/Qiskit/qiskit-optimization/blob/57b59c64bf0ba72057465f2a6ff3560b26239ced/qiskit_optimization/algorithms/minimum_eigen_optimizer.py#L244
https://github.com/Qiskit/qiskit-optimization/blob/57b59c64bf0ba72057465f2a6ff3560b26239ced/qiskit_optimization/algorithms/multistart_optimizer.py#L87
https://github.com/Qiskit/qiskit-optimization/blob/57b59c64bf0ba72057465f2a6ff3560b26239ced/qiskit_optimization/algorithms/slsqp_optimizer.py#L174
https://github.com/Qiskit/qiskit-optimization/blob/57b59c64bf0ba72057465f2a6ff3560b26239ced/qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py#L322
https://github.com/Qiskit/qiskit-optimization/blob/57b59c64bf0ba72057465f2a6ff3560b26239ced/qiskit_optimization/algorithms/grover_optimizer.py#L163
https://github.com/Qiskit/qiskit-optimization/blob/57b59c64bf0ba72057465f2a6ff3560b26239ced/qiskit_optimization/algorithms/admm_optimizer.py#L403

tomtuamnuq:
### Summary
Adds Maximize To Minimize converter

Fixes #41

### Details and comments
Introduces a new converter class ``MaximizeToMinimize`` for ``QuadraticProgram`` as described in #41 .
It converts a problem to a minimization problem. The converter is in the default converters list of ``OptimizationAlgorithm``. Algorithms and ``QuadraticProgramToQubo`` are now using ``MaximizeToMinimize``.


qiskit_optimization/algorithms/admm_optimizer.py
```diff
@@ -290,7 +289,9 @@ def solve(self, problem: QuadraticProgram) -> ADMMOptimizationResult:
         problem = int2bin.convert(problem)
 
         # we deal with minimization in the optimizer, so turn the problem to minimization
-        problem, sense = self._turn_to_minimization(problem)
+        from ..converters.maximize_to_minimize import MaximizeToMinimize
```
woodsp-ibm, line 292: Is there any problem with having the import at the head of the file with the others. Sometimes we do have these as inline imports later, but its normally to avoid some cyclic import problem that would occur otherwise, if there was no other realistic way to deal with it.

qiskit_optimization/algorithms/cobyla_optimizer.py
```diff
@@ -110,9 +110,15 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
         """
         self._verify_compatibility(problem)
 
+        # we deal with minimization in the optimizer, so turn the problem to minimization
+        from ..converters.maximize_to_minimize import MaximizeToMinimize
```
woodsp-ibm, line 114: Similar comment about inline import - so I will not comment on any more that might exist, suffice to say the above comment on inline import versus being at head of file applies.

qiskit_optimization/converters/maximize_to_minimize.py
```diff
@@ -0,0 +1,78 @@
...
+from typing import Optional, Union, Tuple, List
+
+import numpy as np
+
+from .quadratic_program_converter import QuadraticProgramConverter
+from ..exceptions import QiskitOptimizationError
+from ..problems.quadratic_objective import QuadraticObjective
+from ..problems.quadratic_program import QuadraticProgram
+
+logger = logging.getLogger(__name__)
```
woodsp-ibm, line 26: Minor comment: as there is no logging done in this file we can remove this from the file since its not used.

qiskit_optimization/converters/maximize_to_minimize.py
```diff
@@ -0,0 +1,78 @@
...
+        Args:
+            problem: The problem to be solved, that is a maximization or minimization problem.
+
+        Returns:
+            The converted problem, that is a minimization problem.
+        """
+
+        # Copy original QP as reference.
+        self._src = copy.deepcopy(problem)
+        self._dst = copy.deepcopy(problem)
```
woodsp-ibm, line 48: Is there really a need to store these? In terms of _dst its just returned from convert here. And for _src, the only place I see it used later is in interpret, where instead we could potentially simply just store problem.get_num_vars() rather than a copy of the whole original src QP.

tomtuamnuq:
The installation of cvxpy by just using pip install qiskit-optimization[cvxpy] was not enough to set up the complete test on my local machine and I missed to check whether the tests were still skipped. Sorry for that.

I have checked the failed Unit Test: TestWarmStartQAOAOptimizer.test_max_cut
The problem is, that the presolver GoemansWilliamsonOptimizer is now getting a minimization qubo. Therefore, opt_result in https://github.com/Qiskit/qiskit-optimization/blob/3e21815034c72030eab17f2865895dc0d12076a4/qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py#L294 is now 

optimal function value: 0.0
optimal value: [0 0 0 0]
status: SUCCESS

Instead of
optimal function value: 4.0
optimal value: [0 1 1 0]
status: SUCCESS

This leads to a wrong result.

Maybe we should not add MaximizeToMinimize always to the default converters list in https://github.com/Qiskit/qiskit-optimization/blob/3e21815034c72030eab17f2865895dc0d12076a4/qiskit_optimization/algorithms/optimization_algorithm.py#L354

Then, we can alter the UnitTest to use a fixed List of converters(without MaximizeToMinimize) instead of the default list.

Maybe GoemansWilliamsonOptimizer should give a hint if it is used with a minimization problem?

woodsp-ibm (commented):
Nice job for a first contribution! You have even done a reno yaml file, nice!

qiskit_optimization/algorithms/admm_optimizer.py
tomtuamnuq: The inline imports were added similar to the existing ones a few lines before:
`from ..converters.integer_to_binary import IntegerToBinary` 
I will put the imports of MaximizeToMinimize always to the head of the files, as this works properly.

qiskit_optimization/converters/maximize_to_minimize.py
tomtuamnuq: I was also thinking whether this is needed at all but have chosen an implementation which is consistent with the existing converters. All of them store self._src and self._dst. 
LinearEqualityToPenalty saves self._dst and it is not used later(as far as I know). The same goes for IntegerToBinary: even if no integer variables exist the problem is copied. I thought this is done to avoid side effects or something I don`t see.

If this is not the case, I would just store the number of variables as proposed.

woodsp-ibm: Ok, consistency is good. Maybe @t-imamichi has input here - it just seems to me a bit inefficient storing things like that when not really needed.

qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -376,7 +378,10 @@ def _prepare_converters(converters: Optional[Union[QuadraticProgramConverter,
             return [converters]
         elif isinstance(converters, list) and \
                 all(isinstance(converter, QuadraticProgramConverter) for converter in converters):
-            return converters
+            if any(isinstance(converter, MaximizeToMinimize) for converter in converters):
+                return converters
+            else:
+                return converters + [MaximizeToMinimize()]
```
adekusar-drl, line 384: Do we need a `MaximizeToMinimize` converter in every optimizer by default?
@t-imamichi what do you think?

qiskit_optimization/converters/quadratic_program_to_qubo.py
```diff
@@ -43,19 +43,21 @@ def __init__(self, penalty: Optional[float] = None) -> None:
...
+        from ..converters.maximize_to_minimize import MaximizeToMinimize
 
         self._int_to_bin = IntegerToBinary()
         self._ineq_to_eq = InequalityToEquality(mode='integer')
         self._penalize_lin_eq_constraints = LinearEqualityToPenalty(penalty=penalty)
+        self._max_to_min = MaximizeToMinimize()
 
     def convert(self, problem: QuadraticProgram) -> QuadraticProgram:
-        """Convert a problem with linear equality constraints into new one with a QUBO form.
+        """Convert a problem with linear inequality constraints into new one with a QUBO form.
```
adekusar-drl, line 54: Why is this change here? It is unclear how it is related to the max to min converter.

test/converters/test_converters.py
```diff
@@ -321,6 +323,21 @@ def test_penalize_integer(self):
         new_x = conv.interpret([0, 1, -1])
         np.testing.assert_array_almost_equal(new_x, [0, 1, -1])
 
+    def test_maximize_to_minize(self):
```
adekusar-drl, line 326:
```suggestion
    def test_maximize_to_minimize(self):
```

qiskit_optimization/algorithms/admm_optimizer.py
tomtuamnuq: Fixed imports

qiskit_optimization/algorithms/cobyla_optimizer.py
tomtuamnuq: Fixed imports

qiskit_optimization/converters/maximize_to_minimize.py
tomtuamnuq: Removed the logger

qiskit_optimization/algorithms/multistart_optimizer.py
```diff
@@ -83,18 +86,9 @@ def multi_start_solve(self, minimize: Callable[[np.ndarray], Tuple[np.ndarray, A
...
             logger.debug("minimize done in: %s seconds", str(time.time() - t_0))
 
-            # we minimize, to get actual objective value we must multiply by the sense value
-            fval = problem.objective.evaluate(x) * problem.objective.sense.value
-            # we minimize the objective
-            if fval < fval_sol:
-                # here we get back to the original sense of the problem
-                fval_sol = fval * problem.objective.sense.value
-                x_sol = x
-                rest_sol = rest
```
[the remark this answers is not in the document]
tomtuamnuq: Added the lines. I somehow thought the loop would converge to the minimum by calling ``x, rest = minimize(x_0)`` and the statements after ``if fval < fval_sol:`` were for maximization problems only. Understand it now, thanks for clarification!

qiskit_optimization/converters/quadratic_program_to_qubo.py
tomtuamnuq: I have updated the docstring and thought it would be okay to fix the general description with it. ``QuadraticProgramToQubo`` accepts inequality constraints. Therefore the description was not precise. Should this be done separately?

test/converters/test_converters.py
tomtuamnuq: Fixed with test case extension.

test/converters/test_converters.py
```diff
@@ -321,6 +323,21 @@ def test_penalize_integer(self):
...
+        op.integer_var(name='x{}'.format(2), lowerbound=-3, upperbound=3)
+        op.maximize(constant=3, linear={'x0': 1}, quadratic={('x1', 'x2'): 2})
+        conv = MaximizeToMinimize()
+        op_min = conv.convert(op)
+        self.assertEqual(op_min.objective.sense, op.objective.Sense.MINIMIZE)
+        x = [0, 1, 2]
+        fval_min = op_min.objective.evaluate(conv.interpret(x))
+        self.assertAlmostEqual(fval_min, -7)
+        self.assertAlmostEqual(op.objective.evaluate(x), -fval_min)
+
```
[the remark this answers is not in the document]
tomtuamnuq: Extended the test case to check minimization conversion.

qiskit_optimization/converters/quadratic_program_to_qubo.py
adekusar-drl: Ah, got it, sorry. That'd right, but then it should be mentioned that inequality and linear equality constraints are converted as well integers to binaries. This is what QP_to_QUBO does.

t-imamichi:
Could you apply black code formatter by `make black` as explained in https://github.com/Qiskit/qiskit-optimization/blob/main/CONTRIBUTING.md#pull-request-checklist
We recently apply black to the entire code base. See #117 for details.

t-imamichi:
Sorry. Could format again by `black -l 100`? We added a missing config #118.

qiskit_optimization/algorithms/optimization_algorithm.py
t-imamichi: I don't think we need to append max2min here. But, we need to raise an error if an algorithm without max2min receives maximization problem.

t-imamichi:
I think we need to simplify each algorithm to use both `OptimizationAlgorithm._convert` and `OptimizationAlgorithm._interpret`.

qiskit_optimization/converters/maximize_to_minimize.py
t-imamichi: Yes, some converters may store problems inefficiently. I would like to improve them in the future. @tomtuamnuq can avoid inefficient store here. For example, storing only get_num_vars or a shallow copy `self._src = problem`.

tomtuamnuq:
> 
> 
> Sorry. Could format again by `black -l 100`? We added a missing config #118.

Hello @t-imamichi , 
since I have merged the config file in, prior to commit with "apply make black", it should be okay. 
Running black with line length spec does not change anything.

qiskit_optimization/converters/maximize_to_minimize.py
tomtuamnuq: I`ll only store num_vars. Should converters always return a new instance @t-imamichi in ``convert`` ?  Or can I just return the original problem if it is minimization? That would avoid copying.

qiskit_optimization/algorithms/optimization_algorithm.py
tomtuamnuq: Okay @t-imamichi , I will remove the appending of Max2Min to the list. 
I would propose error raising in the following optimizers:
grover, minimum_eigen_optimizer

We could do this after conversion of the problem in ``solve`` method , e.g. after
https://github.com/Qiskit/qiskit-optimization/blob/724158ba42f4b75758edc1de42f6f665660ec27e/qiskit_optimization/algorithms/grover_optimizer.py#L175-L176

        # convert problem to minimization QUBO problem
        problem_ = self._convert(problem, self._converters)
        if problem_.objective.sense == QuadraticObjective.Sense.MAXIMIZE:
                "Optimizer internally deals with minimization problem only. \
                Please ensure conversion."
            )

What do you think?

There are 2 more optimizers without max2min: recursive_meo, warm_start_qaoa

Recursive meo uses other optimizers and should therefore not throw an error. The error would be thrown eventually by those optimizers solve methods.

Warm_start_qaoa uses an optimizer as pre-solver. Like described in my comment(shortly after PR) this can create conflicts, for example with ``GoemansWilliamsonOptimizer``. We should let the user take care by an appropriate converters list.

t-imamichi: The compatibility is checked by `OptimizationAlgorithm._verify_compatibility` and internally `get_compatibility_msg` returns some message if the problem is not compatible. For GroverOptimizer, it checks the compatibility as follows.
https://github.com/Qiskit/qiskit-optimization/blob/6be880b992d9a2f1de8cc89cfd999b786f98ce8a/qiskit_optimization/algorithms/grover_optimizer.py#L107-L119

If the default converter is used in GroverOptimizer, we don't need to take care of the error handling. But, if a customized converter is given, it gets tricky. Let me think.

t-imamichi: We have to think of the case where a customized converter is given to `GroverOptimizer` in a separate PR because the current PR is already large enough.

I recommend the following to keep this PR simple.
1. not add an extra check of compatibility than `_verify_compatibility` in this PR
2. revert `_prepare_converters` not to append max2min.

qiskit_optimization/converters/maximize_to_minimize.py
t-imamichi: If minimization, you can return the original.

tomtuamnuq: MaximizeToMinimize only stores the number of variables now and avoids making a copy if the original problem is already a minimization problem.

qiskit_optimization/algorithms/optimization_algorithm.py
tomtuamnuq: Converter max2min is not appended to the list anymore and there is no extra check.

qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -393,6 +390,7 @@ def _prepare_converters(
             isinstance(converter, QuadraticProgramConverter) for converter in converters
         ):
             return converters
+
```
t-imamichi, line 393:
```suggestion
```

qiskit_optimization/algorithms/multistart_optimizer.py
```diff
@@ -69,10 +70,16 @@ def multi_start_solve(
         Returns:
             The result of the multi start algorithm applied to the problem.
         """
+
```
t-imamichi, line 73:
```suggestion
```

qiskit_optimization/converters/quadratic_program_to_qubo.py
t-imamichi: It can deal with both equality and inequality constraints. So, it might be good to mention just "linear constraints".

qiskit_optimization/algorithms/optimization_algorithm.py
tomtuamnuq: Fixed

qiskit_optimization/algorithms/multistart_optimizer.py
tomtuamnuq: Fixed

qiskit_optimization/converters/quadratic_program_to_qubo.py
tomtuamnuq: Fixed docstring to just mention linear constraints.

t-imamichi:
It may require #136 to pass unit tests. Please wait for a while.

qiskit_optimization/algorithms/cobyla_optimizer.py
```diff
@@ -117,9 +118,14 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
...
 
+        # we deal with minimization in the optimizer, so turn the problem to minimization
+        max2min = MaximizeToMinimize()
+        original_problem = problem
+        problem = self._convert(problem, max2min)
+
         # construct quadratic objective function
         def objective(x):
-            return problem.objective.sense.value * problem.objective.evaluate(x)
+            return problem.objective.evaluate(x)
```
adekusar-drl, line 128: This could be a lambda, since it is a one liner. But this is a matter of taste, so you can keep it as is.

qiskit_optimization/algorithms/slsqp_optimizer.py
```diff
@@ -183,13 +184,17 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
...
+        problem = self._convert(problem, max2min)
 
         # construct quadratic objective function
         def _objective(x):
-            return problem.objective.sense.value * problem.objective.evaluate(x)
+            return problem.objective.evaluate(x)
 
         def _objective_gradient(x):
-            return problem.objective.sense.value * problem.objective.evaluate_gradient(x)
+            return problem.objective.evaluate_gradient(x)
```
adekusar-drl, line 197: Same here, could be a lambda, I'm not insisting :)

test/algorithms/test_warm_start_qaoa.py
```diff
@@ -52,13 +57,19 @@ def test_max_cut(self):
         backend = BasicAer.get_backend("statevector_simulator")
         qaoa = QAOA(quantum_instance=backend, reps=1)
         aggregator = MeanAggregator()
+        # a list of converters without max2min
+        ineq2eq = InequalityToEquality()
+        int2bin = IntegerToBinary()
+        penalize = LinearEqualityToPenalty()
+        converters = [ineq2eq, int2bin, penalize]
```
adekusar-drl, line 64: I'm not convinced all of the converters are required to pass to the optimizer in this test. Does the test fail without them?

adekusar-drl (commented):
One thing to check out is the warm start QAOA test. Other than that, looks good to me! Thanks a lot!

t-imamichi:
LGTM

qiskit_optimization/algorithms/cobyla_optimizer.py
t-imamichi: We don't need to make a function. It is sufficient to just pass `problem.objective.evaluate` to `fmin_cobyla` as `x = fmin_cobyla(problem.objective.evaluate, ...)`

qiskit_optimization/algorithms/slsqp_optimizer.py
t-imamichi: We can simply remove `_objective` and `_objective_gradient` by calling `output = fmin_slsqp(problem.objective.evaluate, x_0, ..., fprime=problem.objective.evaluate_gradient, ...)`

qiskit_optimization/algorithms/cobyla_optimizer.py
adekusar-drl: Oh yeah, that's rights! Shame on me, there no need even for lambda!

qiskit_optimization/algorithms/slsqp_optimizer.py
adekusar-drl: Agree! Without functions it will look better.

qiskit_optimization/algorithms/cobyla_optimizer.py
tomtuamnuq: It is just passed problem.objective.evaluate now.

qiskit_optimization/algorithms/slsqp_optimizer.py
tomtuamnuq: It is just passed problem.objective.evaluate and problem.objective.evaluate_gradient now.

test/algorithms/test_warm_start_qaoa.py
tomtuamnuq: If a converters list with max2min is passed, the problem is converted to a minimization problem before it is given to the presolver. The same happens if no converters list is given, because ``QuadraticProgramToQubo`` is used in that case.

Since in the test case ``test_max_cut`` the presolver is ``GoemansWilliamsonOptimizer``, it fails. 
``GoemansWilliamsonOptimizer`` expects a maximization problem. The ``solve`` method of the presolver is called with the converted_problem ( Line 309). 

I listed the outcomes of the presolver solve method for this test case in a very early comment on the PR:
> The problem is, that the presolver GoemansWilliamsonOptimizer is now getting a minimization qubo. Therefore, opt_result in
> 
> https://github.com/Qiskit/qiskit-optimization/blob/3e21815034c72030eab17f2865895dc0d12076a4/qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py#L294
> is now
> 
> optimal function value: 0.0
> optimal value: [0 0 0 0]
> status: SUCCESS
> 
> Instead of
> optimal function value: 4.0
> optimal value: [0 1 1 0]
> status: SUCCESS
> 
> This leads to a wrong result.
> 
...
> Then, we can alter the UnitTest to use a fixed List of converters(without MaximizeToMinimize) instead of the default list.
> 
> Maybe GoemansWilliamsonOptimizer should give a hint if it is used with a minimization problem?

adekusar-drl: OK, I see the problem now. I believe the warm start optimizer must work without passing additional converters. So, the initial version of the test should stay unchanged. That means we have a few options:

- Do not add max to min in quadratic_program_to_qubo.py. @t-imamichi I guess this is not what we want to achieve, right?
- Modify Goemans-Williamson in a way it converts max to min, may be with an additional warning. This optimizer is built to solve just a very specific type of problems. Sounds messy, requires an additional min to max converter, but still a viable solution and I can take care of this.
- The easiest, tbh, add a check to the warm start if there's a min problem is passed and presolver is Goemans-Williamson, then raise an error saying something like "please provide converters manually". And then we change the test.

@t-imamichi, what do you think? 

@tomtuamnuq I really appreciate your patience and efforts in this PR. It looked simpler initially.

t-imamichi: I see. My first idea is to include max2min converter in quadratic_to_qubo because all algorithms suppose minimization algorithm. If Goemans-Williamson takes maximization problem, my idea does not work. Let me check the code.

t-imamichi: I agree with @adekusar-drl's idea not to include max2min converter in qp_to_qubo because Goemans-Williamson is a maximization algorithm. We need to manually append max2min in `self._converters` in some algorithms if necessary.

As for Goemans-Williamson, we need to check whether the sense of the problem is maximization or not in `GoemansWilliamsonOptimizer.get_compatibility_msg`.
Another option is that we make min2max converter and include it as part of GW.

@tomtuamnuq I'm sorry I did not notice Goemans-Williamson is a maximization algorithm. I appreciate your patience and contribution to our code.

adekusar-drl: @t-imamichi Are sure we should remove max2min in qp_to_qubo just because of one optimizer (which is a utility one)? I fine with either solution here anyway.

Changes to Goemans-Williamson is a separate PR, I agree.

tomtuamnuq: Hello @adekusar-drl and @t-imamichi , thanks for your feedback on this. I will revert changes in qp_to_qubo. I have checked which algorithms use ``self._prepare_converters`` method: grover, meo, recursive_meo, warm_start_qaoa

grover deals with minimization only. We must make sure max2min is in the list ``self._converters``. 
I would just overwrite the inherited ``_prepare_converters`` method, okay?

warm_start_qaoa(as we have seen) must not use max2min before passing the problem to the presolver.
warm_start_qaoa uses the ``_solve_internal`` method of meo (which assumes a minimization problem with current implementation). 

meo can do maximization problems as well but in that case we need:

https://github.com/Qiskit/qiskit-optimization/blob/93b322f954972dea11a0d4a0957c79d28648e313/qiskit_optimization/algorithms/minimum_eigen_optimizer.py#L222

in method ``_solve_internal`` as before. It should then work.
The same goes for recursive_meo. Since recursive_meo is using other optimizers we should not apply max2min here.

Like discussed before it would be good to do  Goemans-Williamson changes in another PR to keep this PR simple.

t-imamichi: @adekusar-drl If GW is not supposed to use as an independent optimizer, it might be good to adjust GW as a separate PR. Sorry, I didn't understand the usage of GW. If we take care of GW as a separate PR. We can merge this PR as is. Is it correct?

test/algorithms/test_warm_start_qaoa.py
```diff
@@ -52,13 +57,19 @@ def test_max_cut(self):
...
         optimizer = WarmStartQAOAOptimizer(
             pre_solver=presolver,
             relax_for_pre_solver=False,
             qaoa=qaoa,
             epsilon=0.25,
             num_initial_solutions=10,
             aggregator=aggregator,
+            converters=converters,
```
t-imamichi, line 72: Because maxcut does not require any converters, the following should work.
```suggestion
            converters=[]
```

test/algorithms/test_warm_start_qaoa.py
t-imamichi: I notice that we maybe don't need any converters in this case. https://github.com/Qiskit/qiskit-optimization/pull/114/files#r636922553

tomtuamnuq: @t-imamichi I have tested and yes the empty list works.

t-imamichi: Yes, it is because maxcut is a QUBO even without any conversion. Let's wait for a response of @adekusar-drl

adekusar-drl: @t-imamichi GW is a helper for the warm start QAOA. It solves a relaxed problem, actually samples possible solutions, and then a customized QAOA circuit is built from the sampled solutions. So, this is a utility optimizer.

As a temporary solution in this PR `converters=[]` is ok, but I'd expect the warm start optimizer to be correctly working even if I don't pass any converters. And this is not the case if this PR is merged as it is now, since if there's no converters, defaults are applied and max is converted to min and GW fails eventually. This is the only thing makes me worry. I'm ok if we stream-line this behavior separately, in another PR.

t-imamichi, line 64: Let's simplify converters by passing `[]`.
```suggestion
```

test/algorithms/test_warm_start_qaoa.py
```diff
@@ -28,6 +28,11 @@
     MeanAggregator,
     WarmStartQAOAOptimizer,
 )
+from qiskit_optimization.converters import (
+    InequalityToEquality,
+    IntegerToBinary,
+    LinearEqualityToPenalty,
+)
```
t-imamichi, line 35:
```suggestion
```

t-imamichi:
Let's see the CI result. I think this PR is ready if CI passes.

t-imamichi (approved):
Thank you for your cooperation! LGTM.

PR #115, opened 2021-04-29
#110 Raised ValueError if trials <= 0

### LINKED ISSUE #106: Check whether the number of trials is positive or not in `MultiStartOptimizer`

alimustafashah:
Fixes #106


t-imamichi:
Could you write details in the description? You need to link this PR to #106 and #110.

qiskit_optimization/algorithms/multistart_optimizer.py
```diff
@@ -52,6 +52,10 @@ def __init__(self, trials: int = 1, clip: float = 100.) -> None:
                 with the ``-clip`` or ``clip`` values correspondingly for the initial guesses.
         """
         super().__init__()
+        
+        if trials <= 0:
+            raise ValueError('Number of trials should be positive')
```
woodsp-ibm, line 57: How about
```suggestion
            raise ValueError(f'Number of trials should be 1 or higher, but was {trials}')
```
(Or whatever text you prefer such as a minimum of 1 etc) and it can be helpful to see the value that was actually passed, that caused the failure in the msg. Same for below.

I imagine CI will complain since it will want Raises:  added to the docstrings too

alimustafashah: So what do you suggest then?

t-imamichi:
Could you write a unit test that raises `ValueError` if `trials <= 0`?

qiskit_optimization/algorithms/multistart_optimizer.py
woodsp-ibm: It indeed failed you look at the build details failure, plus there is some whitespace at end of line that is checked for - these are the errors
```
qiskit_optimization/algorithms/multistart_optimizer.py:118:89: C0303: Trailing whitespace (trailing-whitespace)
qiskit_optimization/algorithms/multistart_optimizer.py:41:4: W9006: "ValueError" not documented as being raised (missing-raises-doc)
qiskit_optimization/algorithms/multistart_optimizer.py:111:4: W9006: "ValueError" not documented as being raised (missing-raises-doc)
```
As I mentioned above you need to add Raises: to the docstrings of the the methods, here is an example for elsewhere in the optimization code https://github.com/Qiskit/qiskit-optimization/blob/3e21815034c72030eab17f2865895dc0d12076a4/qiskit_optimization/problems/quadratic_program.py#L289-L305

alimustafashah:
> Could you write a unit test that raises `ValueError` if `trials <= 0`?

@t-imamichi can you guide me for the unit testing?

woodsp-ibm:
For unit test, since the cobyla optimizer is a subclass of the multistart optimizer maybe look at that https://github.com/Qiskit/qiskit-optimization/blob/main/test/algorithms/test_cobyla_optimizer.py and the last test there has a valid trials.

So what you would do is pass an invalid value on the constructor as one test and an invalid value on the setter as another. YOu can check they raise a value error like this code tests a certain error it expects will happen https://github.com/Qiskit/qiskit-optimization/blob/92403b6f7c4e067602acc148411da1742fb356e6/test/problems/test_linear_constraint.py#L77-L78 

You might like also to check that the value is still the prior value that was set i.e. that the value was indeed ignored rather than say get set internally and then the error raised. Anyway look at the unit tests, you can see what they do, they basically run the code and assert certain expectations over its behavior - i.e. check it behaves as expected

t-imamichi:
I added a unit test of number trials in tests of Cobyla.

qiskit_optimization/algorithms/multistart_optimizer.py
```diff
@@ -51,8 +51,13 @@ def __init__(self, trials: int = 1, clip: float = 100.0) -> None:
             clip: Clipping parameter for the initial guesses in the multi-start method.
                 If a variable is unbounded then the lower bound and/or upper bound are replaced
                 with the ``-clip`` or ``clip`` values correspondingly for the initial guesses.
+
+        Raises:
+            ValueError: if the variable trials have a value smaller than 1.
```
woodsp-ibm, line 56:
```suggestion
            ValueError: if the variable trials has a value smaller than 1.
```

woodsp-ibm (approved):

qiskit_optimization/algorithms/multistart_optimizer.py
```diff
@@ -51,8 +51,13 @@ def __init__(self, trials: int = 1, clip: float = 100.0) -> None:
             clip: Clipping parameter for the initial guesses in the multi-start method.
                 If a variable is unbounded then the lower bound and/or upper bound are replaced
                 with the ``-clip`` or ``clip`` values correspondingly for the initial guesses.
+
+        Raises:
+            ValueError: if the variable trials has a value smaller than 1.
         """
         super().__init__()
+        if trials <= 0:
```
woodsp-ibm, line 59: I guess it would have been simpler and more match the docstring to have had `if trails < 1` in both places rather than <= 0

t-imamichi: Yes, it makes sense. I update the code and title.

woodsp-ibm (approved):

PR #130, opened 2021-05-13
Increase num qubit for grover bit ordering unit test

t-imamichi:
### Summary

Follow-up of #97 

I notice that a unit test may fail on Linux as follows.
```
======================================================================
FAIL: test_bit_ordering_1_sv (test.algorithms.test_grover_optimizer.TestGroverOptimizer)
Test bit ordering
----------------------------------------------------------------------
  File "/home/ima/envs/dev/lib/python3.8/site-packages/ddt.py", line 182, in wrapper
    return func(self, *args, **kwargs)
  File "/home/ima/qiskit/opt/test/algorithms/test_grover_optimizer.py", line 236, in test_bit_ordering
    self.assertEqual(results.fval, opt_sol)
AssertionError: 0.0 != -2
----------------------------------------------------------------------
```

It is required to make #129 pass.

### DIFF
--- a/test/algorithms/test_grover_optimizer.py
+++ b/test/algorithms/test_grover_optimizer.py
@@ -231,7 +231,7 @@ class TestGroverOptimizer(QiskitOptimizationTestCase):
         op.from_docplex(mdl)
         opt_sol = -2
         success = OptimizationResultStatus.SUCCESS
-        grover_optimizer = GroverOptimizer(3, num_iterations=10, quantum_instance=q_instance)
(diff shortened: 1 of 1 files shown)


woodsp-ibm (approved):

t-imamichi:
We need more investigation because @manoelmarques runs CI again with #129 and we found that CI succeeds without this PR. It looks strange...
https://github.com/Qiskit/qiskit-optimization/runs/2575783380

t-imamichi:
I close it because we implemented a better solution #132.

PR #139, opened 2021-05-19
Revise CplexOptimizer

### LINKED ISSUE #133: Add optimality target parameter to cplex.solve() in CplexOptimizer
When using Cplex to solve a Quadratic problem (or as a presolver in Warm Start)  the user may get "CPLEX Error 5002: objective is not convex". CPLEX has the option to find a solution that satisfies first-order optimality conditions (reaching a potentially local optimum) by setting optimality target to 2. This may be particularly useful in getting a warm start even when the objective is not convex.

Here is the documentation of the parameter:
https://www.ibm.com/docs/en/icos/12.10.0?topic=parameters-optimality-target

The user now can extend CplexOptimizer and override solve adding the following after getting cplex 
        cplex.parameters.optimalitytarget.set(2)
or, we can add an optimality target to the solve method in CplexOptimizer which defaults to 1 (convex optimization) and gives the user the opportunity to change.

t-imamichi:
### Summary

Simplifies `CplexOptimizer` by directly calling `docplex.mp.model.Model.solve`.
Adds `cplex_parameters` to set various CPLEX options.

Fixes #133

### DIFF
--- a/qiskit_optimization/algorithms/cplex_optimizer.py
+++ b/qiskit_optimization/algorithms/cplex_optimizer.py
@@ -18,12 +18,12 @@ from warnings import warn
 
 from qiskit.exceptions import MissingOptionalLibraryError
 
-from ..problems.quadratic_program import QuadraticProgram
 from .optimization_algorithm import (
     OptimizationAlgorithm,
     OptimizationResult,
     OptimizationResultStatus,
 )
+from ..problems.quadratic_program import QuadraticProgram
 
 logger = logging.getLogger(__name__)
 
@@ -70,7 +70,7 @@ class CplexOptimizer(OptimizationAlgorithm):
             )
 
         self._disp = disp
-        self._cplex_parameters = {} if cplex_parameters is None else cplex_parameters
+        self._cplex_parameters = cplex_parameters
 
     @staticmethod
     def is_cplex_installed():
@@ -95,12 +95,12 @@ class CplexOptimizer(OptimizationAlgorithm):
         self._disp = disp
 
     @property
-    def cplex_parameters(self) -> Dict[str, Any]:
+    def cplex_parameters(self) -> Optional[Dict[str, Any]]:
         """Returns parameters for CPLEX"""
         return self._cplex_parameters
 
     @cplex_parameters.setter
-    def cplex_parameters(self, parameters: Dict[str, Any]):
+    def cplex_parameters(self, parameters: Optional[Dict[str, Any]]):
         """Set parameters for CPLEX
         Args:
             parameters: The parameters for CPLEX
@@ -142,9 +142,10 @@ class CplexOptimizer(OptimizationAlgorithm):
         mod = problem.to_docplex()
         sol = mod.solve(log_output=self._disp, cplex_parameters=self._cplex_parameters)
         if sol is None:
+            # no solution is found
             warn("CPLEX cannot solve the model")
             x = [0.0] * mod.number_of_variables
-            result = OptimizationResult(
+            return OptimizationResult(
                 x=x,
                 fval=problem.objective.evaluate(x),
                 variables=problem.variables,
@@ -152,12 +153,12 @@ class CplexOptimizer(OptimizationAlgorithm):
                 raw_results=None,
             )
         else:
+            # a solution is found
             x = sol.get_values(mod.iter_variables())
-            result = OptimizationResult(
+            return OptimizationResult(
                 x=x,
                 fval=sol.get_objective_value(),
                 variables=problem.variables,
                 status=self._get_feasibility_status(problem, x),
                 raw_results=sol,
             )
-        return result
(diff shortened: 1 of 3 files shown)


qiskit_optimization/algorithms/cplex_optimizer.py
```diff
@@ -84,6 +94,19 @@ def disp(self, disp: bool):
         """
         self._disp = disp
 
+    @property
+    def cplex_parameters(self) -> Dict[str, Any]:
+        """Returns parameters for CPLEX"""
+        return self._cplex_parameters
+
+    @cplex_parameters.setter
+    def cplex_parameters(self, parameters: Dict[str, Any]):
```
adekusar-drl, line 103: I guess parameters should be the same as in the constructor, `Optional[Dict[str, Any]]` in both getter/setter?

qiskit_optimization/algorithms/cplex_optimizer.py
```diff
@@ -116,33 +139,25 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
...
+            )
+        else:
+            x = sol.get_values(mod.iter_variables())
+            result = OptimizationResult(
+                x=x,
+                fval=sol.get_objective_value(),
+                variables=problem.variables,
+                status=self._get_feasibility_status(problem, x),
+                raw_results=sol,
+            )
```
adekusar-drl, line 162: May be the code would look better if you pulled the `result` creation out of the `if` statement? This is just an idea, not a suggestion.

qiskit_optimization/algorithms/cplex_optimizer.py
```diff
@@ -116,33 +139,25 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
...
-            x=sol.get_values(),
-            fval=sol.get_objective_value(),
-            variables=problem.variables,
-            status=self._get_feasibility_status(problem, sol.get_values()),
-            raw_results=sol,
-        )
-
-        # return solution
+        mod = problem.to_docplex()
+        sol = mod.solve(log_output=self._disp, cplex_parameters=self._cplex_parameters)
```
adekusar-drl, line 143: Don't really get, why is this better rather than calling cplex directly? Just curious...

qiskit_optimization/algorithms/cplex_optimizer.py
t-imamichi: Because we cannot use the empty dict `{}` as a default value of the method, I used `Optional`. But, setter/getter does not need to use the default value as parameter. So, I prefer without `Optional`.

qiskit_optimization/algorithms/cplex_optimizer.py
t-imamichi: I think there are two reasons
1. docplex is more pythonic than `cplex` and easier to maintain.
2. we need to refer only docplex manual to maintain this code. Previously we need to refer both manuals of cplex and docplex.
Basically, I prefer to hide `cplex`

qiskit_optimization/algorithms/cplex_optimizer.py
t-imamichi: It might be difficult to simplify it because the values of parameters are different in `if` statement. How about returning directly `OptimizationResult`, i.e., `return OptimizationResult{...}` in both `if` and `else` statements.

qiskit_optimization/algorithms/cplex_optimizer.py
```diff
@@ -61,6 +70,7 @@ def __init__(self, disp: bool = False) -> None:
             )
 
         self._disp = disp
+        self._cplex_parameters = {} if cplex_parameters is None else cplex_parameters
```
adekusar-drl, line 73: By the way, this can be expressed by `self._cplex_parameters = cplex_parameters or {}`

qiskit_optimization/algorithms/cplex_optimizer.py
adekusar-drl: I see, usually we have same type hints in properties as in the constructor. I'd rather add a check for `None` in the setter anyway if `None` value may cause an error later.

qiskit_optimization/algorithms/cplex_optimizer.py
adekusar-drl: OK, thanks

qiskit_optimization/algorithms/cplex_optimizer.py
adekusar-drl: OK, fine with me!

qiskit_optimization/algorithms/cplex_optimizer.py
t-imamichi: It's OK to raise a ValueError if None is given.

t-imamichi: I checked the manual again and noticed that `cplex_parameters` is optional. So, `None` is treated as no option. So, I change it to accept `None`.

adekusar-drl (approved):
Unfortunately, my approval won't merge the code.

manoelmarques (approved):

t-imamichi:
Thank you @manoelmarques @adekusar-drl !

PR #140, opened 2021-05-20
Raise an appropriate error if an empty objective function is used

t-imamichi:
### Summary

An error occurs if no objective function is set. This PR fixes it.

```python

qp = QuadraticProgram()
qp.binary_var_list(3)
print(qp.objective.evaluate([0, 0, 0]))
```

```
# before
  File "obj2.py", line 5, in <module>
    print(qp.objective.evaluate([0, 0, 0]))
  File "/Users/ima/envs/qiskit38/lib/python3.8/site-packages/qiskit_optimization/problems/quadratic_objective.py", line 151, in evaluate
    return self.constant + self.linear.evaluate(x) + self.quadratic.evaluate(x)
  File "/Users/ima/envs/qiskit38/lib/python3.8/site-packages/qiskit_optimization/problems/linear_expression.py", line 151, in evaluate
    val = (x @ self.coefficients.transpose())[0, 0]
  File "/Users/ima/envs/qiskit38/lib/python3.8/site-packages/scipy/sparse/base.py", line 560, in __matmul__
    return self.__mul__(other)
  File "/Users/ima/envs/qiskit38/lib/python3.8/site-packages/cvxpy/interface/scipy_wrapper.py", line 31, in new_method
    return method(self, other)
  File "/Users/ima/envs/qiskit38/lib/python3.8/site-packages/scipy/sparse/base.py", line 479, in __mul__
    raise ValueError('dimension mismatch')
ValueError: dimension mismatch
```

```
# after
  File "obj2.py", line 7, in <module>
    print(qp.objective.evaluate([0, 0, 0]))
  File "/Users/ima/tasks/1_2021/qiskit/opt/qiskit_optimization/problems/quadratic_objective.py", line 173, in evaluate
qiskit_optimization.exceptions.QiskitOptimizationError: 'The shape of the objective function does not match with the number of variables. Need to define the objective function after defining all variables'
```


adekusar-drl (commented):
Could it be just easier to raise an exception saying no objective is defined if `evaluate` is called on a problem without it? It seems more meaningful rather than returning `0.0`. This is sort of misleading, I think.

t-imamichi:
Thanks. Your idea makes sense. Let me try it.

t-imamichi:
The unit test of objective assumes that the objective is not zero if not set.
https://github.com/Qiskit/qiskit-optimization/blob/93b322f954972dea11a0d4a0957c79d28648e313/test/problems/test_quadratic_objective.py#L26-L36

There are two options.
1. raise an error if users does not set objective and access it. We need to modify the unit tests of objective too.
2. adjust shape of expressions in the objective as I did in this PR.

@adekusar-drl Which do you prefer? I'm ok to make the access of objective error if it is not set.

t-imamichi:
Thank you for your idea. I simplified it by raising an error if no obj is set.

qiskit_optimization/problems/quadratic_objective.py
```diff
@@ -162,7 +163,17 @@ def evaluate(self, x: Union[ndarray, List, Dict[Union[int, str], float]]) -> flo
...
+        Raises:
+            QiskitOptimizationError: if the shape of the objective function does not match with
+                the number of variables.
         """
+        n = self.quadratic_program.get_num_vars()
+        if self.linear.coefficients.shape != (1, n) or self.quadratic.coefficients.shape != (n, n):
+            raise QiskitOptimizationError(
+                "The shape of the objective function does not match with the number of variables. "
+                "Need to define the objective function after defining all variables"
+            )
```
adekusar-drl, line 176: Just curious, what if I have a variable that is only in a constraint, but not in objective. Would it work? I guess it should.

t-imamichi: Yes, it works fine with this PR too.
```python

qp = QuadraticProgram()
qp.binary_var_list(3)
qp.linear_constraint({0: 1, 1: 1, 2: 1}, '==', 1)
print(qp)
print(qp.objective.evaluate([0, 0, 0]))
```
```
# output
\ ENCODING=ISO-8859-1
\Problem name: CPLEX

Minimize
 obj:
Subject To
 c0: x0 + x1 + x2 = 1

Bounds
 0 <= x0 <= 1
 0 <= x1 <= 1
 0 <= x2 <= 1

Binaries
 x0 x1 x2
End

  File "obj2.py", line 9, in <module>
    print(qp.objective.evaluate([0, 0, 0]))
  File "/Users/ima/tasks/1_2021/qiskit/opt/qiskit_optimization/problems/quadratic_objective.py", line 173, in evaluate
qiskit_optimization.exceptions.QiskitOptimizationError: 'The shape of the objective function does not match with the number of variables. Need to define the objective function after defining all variables'
```

adekusar-drl: I mean slightly different. Suppose we have an objective, but not all variables are present in the objective. Some of them are only in the constraints for this problem.

t-imamichi: Do you mean this?
```python

qp = QuadraticProgram()
qp.binary_var_list(3)
qp.minimize(linear={0: 1})
qp.linear_constraint({0: 1, 1: 1, 2: 1}, '==', 1)
print(qp)
print(qp.objective.evaluate([0, 0, 0]))
```
```
# output
\ ENCODING=ISO-8859-1
\Problem name: CPLEX

Minimize
 obj: x0
Subject To
 c0: x0 + x1 + x2 = 1

Bounds
 0 <= x0 <= 1
 0 <= x1 <= 1
 0 <= x2 <= 1

Binaries
 x0 x1 x2
End

0.0
```

adekusar-drl: Yes, thanks a lot!

adekusar-drl (approved):
Looks good!

t-imamichi:
Thank you!

adekusar-drl:
ping @manoelmarques for actual approval :)

woodsp-ibm (approved):

woodsp-ibm:
Just a question - do we really want/need to backport this. It seems to be as much an improvement to detect a mismatch between obj fn and vars that a user can easily avoid at present - ie its not really a show stopping/critical bug in stable that is more the criteria for backporting.

t-imamichi:
Thank you for your feedback. You are right. It is not a critical bug.

PR #141, opened 2021-05-20
Add Getting Started Documentation

manoelmarques:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,14 @@ To do this follow the instructions in the
 
 ### Optional Installs
 
-* **IBM CPLEX** may be installed using `pip install 'qiskit-optimization[cplex]'` to enable the reading of `LP` files.
+* **IBM CPLEX** may be installed using `pip install 'qiskit-optimization[cplex]'` to enable the reading of `LP` files and the usage of
+  the `CplexOptimizer`, wrapper for ``cplex.Cplex``.
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -41,6 +41,12 @@ To do this follow the instructions in the
 
 * **IBM CPLEX** may be installed using `pip install 'qiskit-optimization[cplex]'` to enable the reading of `LP` files.
```
woodsp-ibm, line 42: I think we should add here too the 'CplexOptimzer' which is built to use Cplex.

README.md
```diff
@@ -41,6 +41,12 @@ To do this follow the instructions in the
...
 
+* **CVXPY** may be installed using command `pip install 'qiskit-optimization[cvx]'` to install the
+  package. CVXPY being installed will enable the usage of the Goemans-Williamson algorithm as an optimizer `GoemansWilliamsonOptimizer`.
+
+* **Matplotlib** may be installed using command `pip install 'qiskit-optimization[matplotlib]'` to install the
+  package. Matplotlib being installed will enable the usage of graph optimization application classes.
```
woodsp-ibm, line 48: I believe Matplotlib is needed only of you want to 'draw()' the problem/result since its needed by networkx draw method that these use when drawing. The classes can be used otherwise without matplotlib.

PR #142, opened 2021-05-21
Merge spelling dictionaries into one single .pylintdict file

manoelmarques:
### Summary
This merges both dictionaries into a single .pylintdict with all lowercase words.

### Details and comments
Since the Sphinx spell checker is case sensitive, I added a filter to change the word to lowercase in order to check.

To simplify the single initial creation of a merged dictionary, a tool called generate_spell_dict was created. It runs both spell checks and creates a .pylintdict file with the sorted merged content of both dicts in lowercase.


.pylintdict
```diff
@@ -1,41 +1,57 @@
...
 aer's
 al
 annealers
+applegate
 args
+arxiv
 autosummary
 backend
-bb
+behaviour
```
woodsp-ibm, line 13: This one in the Beitish English spelling - since we are using US English it would be preferable to be removed and fixed at source to be `behavior`

manoelmarques: It is fixed

t-imamichi (approved):
Thank you. This is a very convenient tool!

PR #151, opened 2021-05-31
Added support for an indicator constraint to QuadraticProgram

### LINKED ISSUE #13: Add support for docplex.mp.constr.IndicatorConstraint
It would be nice if QuadraticProgram.from_docplex() supports docplex.mp.constr.IndicatorConstraint.

a-matsuo:
### Summary
resolve #13
The PR adds support for an [indicator constraint](http://ibmdecisionoptimization.github.io/docplex-doc/mp/docplex.mp.model.html#docplex.mp.model.Model.add_indicator) to `QuadraticProgram`. 

### Details and comments
An indicator constraint control  the satisfaction of a linear constraint by the value of a binary variable. If the binary variable equals the active value, then the constraint must be satisfied, but otherwise the constraint might or might not be satisfied.
e.g. `x = 1 -> y + z = 1`
- [x] to_docplex
- [x] from_docplex
- [x] converter
- [x] unittests
  - [x] indicator to inequality converter
  - [x] indicator constraint


qiskit_optimization/problems/quadratic_program.py
```diff
@@ -1025,6 +1144,40 @@ def from_docplex(self, model: Model) -> None:
             else:
                 raise QiskitOptimizationError("Unsupported constraint sense: {}".format(constraint))
 
+        # get indicator constraints
+        for constraint in model.iter_logical_constraints():
```
t-imamichi, line 1148: Isn't it `iter_indicator_constraints`? https://ibmdecisionoptimization.github.io/docplex-doc/mp/docplex.mp.model.html?highlight=iter_#docplex.mp.model.Model.iter_indicator_constraints

qiskit_optimization/problems/indicator_constraint.py
```diff
@@ -0,0 +1,131 @@
...
+    def evaluate(self, x: Union[ndarray, List, Dict[Union[int, str], float]]) -> float:
+        """Evaluate the left-hand-side of the constraint.
+
+        Args:
+            x: The values of the variables to be evaluated.
+
+        Returns:
+            The left-hand-side of the constraint given the variable values.
+        """
+        return self.linear.evaluate(x)
```
t-imamichi, line 131: I think you need to check the active value here.

qiskit_optimization/converters/indicator_to_inequality.py
```diff
@@ -0,0 +1,179 @@
...
+
+        Returns:
+            The converted problem, that contain only inequality constraints.
+
+        Raises:
+            QiskitOptimizationError: If a variable type is not supported.
+            QiskitOptimizationError: If an unsupported mode is selected.
+            QiskitOptimizationError: If an unsupported sense is specified.
+        """
+        self._src = copy.deepcopy(problem)
```
ikkoham, line 62: Just a question. Why is this `deepcopy` needed?

t-imamichi: It used to be necessary. Matsuo-san can remove it.

qiskit_optimization/problems/quadratic_program.py
a-matsuo: `iter_logical_constraints()` works since it's the super class of `IndicatorConstraint`, but I think `iter_indicator_constraint` is more specific and more appropriate. I fixed it.

qiskit_optimization/problems/indicator_constraint.py
a-matsuo: I added `evaluate_indicator` to check whether an indicator binary variable equals to `active_value` or not. Feasibility of indicator constraints is checked in `QuadraticProgram.get_feasibility_info()`

a-matsuo:
I think the PR is ready for review. Could you please review this?

t-imamichi:
Thanks. Could you write a reno?

a-matsuo:
Sure (I had written reno but I had forgotten to add it).

qiskit_optimization/converters/indicator_to_inequality.py
```diff
@@ -0,0 +1,175 @@
...
+            if x.vartype == Variable.Type.BINARY:
+                self._dst.binary_var(name=x.name)
+            elif x.vartype == Variable.Type.INTEGER:
+                self._dst.integer_var(name=x.name, lowerbound=x.lowerbound, upperbound=x.upperbound)
+            elif x.vartype == Variable.Type.CONTINUOUS:
+                self._dst.continuous_var(
+                    name=x.name, lowerbound=x.lowerbound, upperbound=x.upperbound
+                )
+            else:
+                raise QiskitOptimizationError("Unsupported variable type {}".format(x.vartype))
```
woodsp-ibm, line 72: I think generally now we have gone over to using f Strings now, since we dropped 3.5 Python support, rather than using this explicit .format() syntax. Not that its incorrect of course to do this, just that going forwards I thin it might be preferable to stick to f Strings so its all done one way.

qiskit_optimization/problems/quadratic_program.py
```diff
@@ -105,6 +107,9 @@ def __init__(self, name: str = "") -> None:
         self._quadratic_constraints = []  # type: List[QuadraticConstraint]
         self._quadratic_constraints_index = {}  # type: Dict[str, int]
 
+        self._indicator_constraints = []  # type: List[IndicatorConstraint]
```
woodsp-ibm, line 110: Just another note going forwards. Since we dropped Python 3.5 support, where the typehints had to be done as comments, since 3.6 in the minimum now we can do typehints directly on the members so we don't have to do it via comments anymore.

test/problems/test_indicator_constraint.py
```diff
@@ -0,0 +1,114 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+""" Test LinearConstraint """
```
woodsp-ibm, line 13:
```suggestion
""" Test IndicatorConstraint """
```

qiskit_optimization/converters/indicator_to_inequality.py
```diff
@@ -0,0 +1,175 @@
...
+        Returns:
+            The result of the original problem.
+
+        Raises:
+            QiskitOptimizationError: if the number of variables in the result differs from
+                                     that of the original problem.
+        """
+        if len(x) != self._src_num_var:
+            raise QiskitOptimizationError(
+                "The number of variables in the passed result differs from "
```
woodsp-ibm, line 172: Might be nice to include the different variable counts in the message.

qiskit_optimization/problems/indicator_constraint.py
```diff
@@ -0,0 +1,181 @@
...
+        if isinstance(x, (list, ndarray)):
+            val = x[index]
+        elif isinstance(x, dict):
+            for ind, value in x.items():
+                if isinstance(ind, str):
+                    ind = self.quadratic_program.variables_index[ind]
+                if index == ind:
+                    val = value
+        else:
+            raise QiskitOptimizationError("Unsupported format for x.")
```
woodsp-ibm, line 179: Should the message, since its more related to type, not format, of x be more like 
`f"Unsupported type '{type(x)}' for x."` so it says type and we include the type in the message

releasenotes/notes/ineq_to_pen-229bc20575b2f483.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
```
woodsp-ibm, line 2: I assume the above reno is correct - this one looks like it should be removed

woodsp-ibm:
Should we be adding a new tutorial, or updating an existing one, to show this new capability. For significant new features I think it would be good, since we have the tutorials now in this same repo, that they are updated/improved at the same time if the new/updated feature warrants it.

a-matsuo:
@woodsp-ibm Thank you for your helpful comments. I fixed the code. I haven't fixed all the f string and typehints, but I fixed them at least in this code.

qiskit_optimization/converters/indicator_to_inequality.py
```diff
@@ -0,0 +1,176 @@
...
+# (C) Copyright IBM 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+"""The inequality to equality converter."""
```
t-imamichi, line 12: Need to rename it

qiskit_optimization/converters/indicator_to_inequality.py
```diff
@@ -0,0 +1,176 @@
...
+
+from .quadratic_program_converter import QuadraticProgramConverter
+from ..exceptions import QiskitOptimizationError
+from ..problems.constraint import Constraint
+from ..problems.quadratic_objective import QuadraticObjective
+from ..problems.quadratic_program import QuadraticProgram
+from ..problems.variable import Variable
+
+
+class IndicatorToInequality(QuadraticProgramConverter):
```
t-imamichi, line 26: I suggest `IndicatorToLinearConstraint` because this converts not only inequality constraints but also equality constraints.

qiskit_optimization/problems/quadratic_program.py
```diff
@@ -793,6 +801,97 @@ def get_num_quadratic_constraints(self) -> int:
...
+        self,
+        binary_var: Variable,
+        linear: Union[ndarray, spmatrix, List[float], Dict[Union[str, int], float]],
+        sense: Union[str, ConstraintSense] = "<=",
+        rhs: float = 0.0,
+        active_value: int = 1,
+        name: Optional[str] = None,
+    ) -> IndicatorConstraint:
+        """Adds an indicator constraint to the quadratic program of the form:
+            z=f -> linear * x sense rhs
```
t-imamichi, line 832: z -> binary_var
f -> active_value

qiskit_optimization/converters/indicator_to_inequality.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def _convert_indicator_constraint(self, problem, indicator_const):
+        # convert indicator constraints to inequality constraints
+        new_linear = indicator_const.linear.to_dict(use_name=True)
+        new_rhs = indicator_const.rhs
+        sense = indicator_const.sense
+        new_name = indicator_const.name + self._delimiter + "indicator"
+        if sense == Constraint.Sense.LE:
+            _, lhs_ub = self._calc_linear_bounds(problem, new_linear)
+            big_m = lhs_ub - new_rhs
```
t-imamichi, line 115: make sure that big_m is a positive value.

a-matsuo: I think it does work even when big_m is a negative value in this case. When it's inactive, it becomes lhs <= the upperbounds of lhs. (When rhs is larger than the upper bound of lhs, the constraint doesn't make sense though).

t-imamichi:
The indicator constraints will be treated as part of a docplex translator #122

PR #158, opened 2021-06-09
Add types-dataclasses for mypy-0.900

t-imamichi:
### Summary

Mypy 0.900, released recently, requires `types-dataclasses`. Mypy shows the following message with Python 3.6.
This PR fixes requirements.txt.

```
qiskit_optimization/algorithms/optimization_algorithm.py:16: error: Library stubs not installed for "dataclasses" (or incompatible with Python 3.6)
qiskit_optimization/algorithms/optimization_algorithm.py:16: note: Hint: "python3 -m pip install types-dataclasses"
qiskit_optimization/algorithms/optimization_algorithm.py:16: note: (or run "mypy --install-types" to install all missing stub packages)
qiskit_optimization/algorithms/optimization_algorithm.py:16: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -19,3 +19,4 @@ mypy>=0.780
 mypy-extensions>=0.4.3
 nbsphinx
 qiskit_sphinx_theme
+types-dataclasses; python_version < '3.7'


manoelmarques:
@t-imamichi thanks fir fixing it. Could you please instead of using requirements.txt, add the types to requirements-dev.txt like it was done for qiskit-finance ? The reason is because the types library is just used by the tool mypy during dev. time and not required for the qiskit-optimization run.

t-imamichi:
Thank you for your feedback. I think it's good to use same version of mypy for all applications and I notice that nature pins `mypy<0.900`. I'm OK to pin it in optimization too. What do you think?

manoelmarques (approved):

manoelmarques:
@t-imamichi I pinned qiskit-nature temporarily because I didn’t have time yesterday to fix the new mypy errors there. I intend to unpin it and fix the errors on the next PR. What you have now is fine. Mypy has a flag that installs the types automatically but it is crashing at the moment. There is an issue against Mypy for this. Once they fix it, I intend to go back to all repos, add the flag and remove the type installations from requirements-dev.txt

PR #163, opened 2021-06-15
Add indicator constraint support in `from_docplex_mp` translator

### LINKED ISSUE #13: Add support for docplex.mp.constr.IndicatorConstraint

t-imamichi:
### Summary

Adds support of indicator constraints in `from_docplex_mp`.
This implementation is based on #151.

Fixes #13
Closes #151

### DIFF
--- a/qiskit_optimization/translators/docplex_mp.py
+++ b/qiskit_optimization/translators/docplex_mp.py
@@ -234,10 +234,10 @@ class _FromDocplexMp:
         binary_var = constraint.binary_var
         active_value = constraint.active_value
         linear_constraint = constraint.linear_constraint
-        linear, sense, rhs = _FromDocplexMp._linear_constraint(var_names, linear_constraint)
-        linear_lb, linear_ub = _FromDocplexMp._linear_bounds(var_bounds, linear)
+        linear, sense, rhs = cls._linear_constraint(var_names, linear_constraint)
+        linear_lb, linear_ub = cls._linear_bounds(var_bounds, linear)
         if sense == "<=":
-            big_m = indicator_big_m or max(0.0, linear_ub - rhs)
+            big_m = max(0.0, linear_ub - rhs) if indicator_big_m is None else indicator_big_m
             if active_value:
                 linear[binary_var.name] = big_m
                 rhs += big_m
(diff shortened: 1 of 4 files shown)


qiskit_optimization/translators/docplex_mp.py
```diff
@@ -119,23 +120,172 @@ def to_docplex_mp(quadratic_program: "QuadraticProgram") -> Model:
...
+                rhs -= big_m
+            else:
+                linear[binary_var.name] = big_m
+            return [(linear, sense, rhs, name)]
+        elif sense == "==":
+            # for equality constraints, add both GE and LE constraints.
+            # linear2, rhs2, and big_m2 are for the GE constraint.
+            linear2 = linear.copy()
+            rhs2 = rhs
+            big_m = indicator_big_m or max(0.0, linear_ub - rhs)
```
woodsp-ibm, line 260: If I supply big M as a value of 0, which is not None - then this logic will not use that value the user supplied. Maybe its not a useful value anyway to supply anyway but if it was the intent to compute one when the user supplied any value other than None then this does not quite do that, Referring to its description
```
indicator_big_m: The big-M value for indicator constraints.
    If ``None``, it is automatically derived from the model.
```

qiskit_optimization/translators/docplex_mp.py
```diff
@@ -119,23 +120,172 @@ def to_docplex_mp(quadratic_program: "QuadraticProgram") -> Model:
...
 
     Note that this supports only basic functions of docplex as follows:
     - quadratic objective function
-    - linear / quadratic constraints
+    - linear / quadratic / indicator constraints
     - binary / integer / continuous variables
 
     Args:
         model: The docplex.mp model to be loaded.
+        indicator_big_m: The big-M value for indicator constraints.
```
woodsp-ibm, line 287: Is this formulation well enough known that it does not require anything further e.g. an example of an indicator contraint in the docstring here being converted so it can be seen where M comes into play.

qiskit_optimization/translators/docplex_mp.py
t-imamichi: Thanks. I revised it as follows
```python
big_m = max(0.0, linear_ub - rhs) if indicator_big_m is None else indicator_big_m
```

qiskit_optimization/translators/docplex_mp.py
t-imamichi: Yes, the big-M formulation is a well known technique. I slightly revised the text to clarify it.

woodsp-ibm (approved):

a-matsuo (approved):
LGTM!

t-imamichi:
I merged #170 and resolved conflicts

woodsp-ibm (approved):

PR #165, opened 2021-06-16
Separate `SubstituteVariable` out from `QuadraticProgram`

t-imamichi:
### Summary

Separeted `SubstituteVariable` out from `QuadraticProgram` to simplify the very long code of `QuadraticProgram`.

### DIFF
--- a/qiskit_optimization/problems/quadratic_program.py
+++ b/qiskit_optimization/problems/quadratic_program.py
@@ -13,32 +13,29 @@
 """Quadratic Program."""
 
 import logging
-from collections import defaultdict
 from collections.abc import Sequence
 from enum import Enum
-from math import fsum, isclose
+from math import isclose
 from typing import Dict, List, Optional, Tuple, Union, cast
 
 import numpy as np
 from docplex.mp.model import Model
 from docplex.mp.model_reader import ModelReader
 from numpy import ndarray, zeros
-from scipy.sparse import spmatrix
-
 from qiskit.exceptions import MissingOptionalLibraryError
 from qiskit.opflow import I, ListOp, OperatorBase, PauliOp, PauliSumOp, SummedOp
 from qiskit.quantum_info import Pauli
+from scipy.sparse import spmatrix
 
-from ..exceptions import QiskitOptimizationError
-from ..infinity import INFINITY
 from .constraint import Constraint, ConstraintSense
 from .linear_constraint import LinearConstraint
-from .linear_expression import LinearExpression
 from .quadratic_constraint import QuadraticConstraint
-from .quadratic_expression import QuadraticExpression
 from .quadratic_objective import QuadraticObjective
+from .substitute_variables import SubstituteVariables
 from .variable import Variable, VarType
 from ..deprecation import DeprecatedType, deprecate_method
+from ..exceptions import QiskitOptimizationError
+from ..infinity import INFINITY
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


qiskit_optimization/problems/substitute_variables.py
```diff
@@ -0,0 +1,308 @@
...
+
+
+class SubstituteVariables:
+    """A class to substitute variables of an optimization problem with constants for other
+    variables"""
+
+    CONST = "__CONSTANT__"
+
+    def __init__(self):
+        self._src = None  # type: Optional[QuadraticProgram]
```
woodsp-ibm, line 37: I don't know if you want to consider changing these while you are doing this. Since we no longer support 3.5 we can put these as proper typehints on a the variables rather than using comments. Something perhaps we can switch over as we change code and see this.

qiskit_optimization/problems/substitute_variables.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def _subs_dict(self, constants, variables):
+        # guarantee that there is no overlap between variables to be replaced and combine input
+        subs = {}  # type: Dict[Union[int, str], Tuple[str, float]]
+        if constants is not None:
+            for i, v in constants.items():
+                # substitute i <- v
+                i_2 = self._src.get_variable(i).name
+                if i_2 in subs:
+                    raise QiskitOptimizationError(
```
woodsp-ibm, line 119: Again something we can consider changing, if you have time/inclination, over to using f-Strings, which we are trying to standardize on using rather than the older .format()

qiskit_optimization/problems/quadratic_program.py
```diff
@@ -998,6 +994,9 @@ def substitute_variables(
                 - Same variable is substituted multiple times.
                 - Coefficient of variable substitution is zero.
         """
+        # pylint: disable=cyclic-import
+        from .substitute_variables import SubstituteVariables
+
         return SubstituteVariables().substitute_variables(self, constants, variables)
```
woodsp-ibm, line 1000: Given the usage here I guess it begs the question why a class rather than a function. The only reason I can see is that it avoids passing src/dst/subs to routines that would need it rather they are stored in the class. I am not suggesting to change things now, since it all works and there are other things todo. More just noting this since I noticed this while reviewing things.

woodsp-ibm (approved):
I made some comments - but since this is just simply a light refactoring of the existing code I'll approve it and leave the option to you to decide whether to go ahead with this as-is or not.

t-imamichi:
Thank you for your review. I refactor it following your comments.

woodsp-ibm:
Should we have a separate unit test for the SubstituteVariable given that it has been separated out now. I was going to look at the coverage to see what ends up being tested via its use by QuadraticProgram but the build did not get that far. But whatever that is it might be good to have a dedicated unit test for the class.

qiskit_optimization/problems/quadratic_program.py
t-imamichi: I will make it a function

qiskit_optimization/problems/substitute_variables.py
t-imamichi: Sure. I will replace them with f-strings

t-imamichi:
I refactored the internal of `substitute_variables`. I separated it out from QP, but I named the new function by `_substitute_variables` as a private function. I'm wondering whether it's good to make another public function of `substitute_variables` on top of the existing method `QP.substitute_variables`.

I think it's a good idea to make a dedicated unit test for the new `_substitute_variables`. If we leave it as a private function, may I make unit tests for a private function?

woodsp-ibm:
You can always leave the substitute_variables new method without the leading underscore. I am sure if I look in my IDE it will be flagging that a private method in another module is being used. The function does not have to go into the init file/ALL so it logically stays within the package. Normally what is tested is the public interface of a module - ie what we expect external users to use/call. I think that would be true of that method however you leave it - but I think all of the other aspects of that module really are expected to be private - i.e. not used by anything other than that module. By having the new method also marked private its kinda hard then to distinguish things.

t-imamichi:
Thank you for your feedback. I leave `substitute_variables` as a function and made a separate unit test, which is currently almost equivalent to the unit test for `QuadraticProgram.suibstitute_variables`.

woodsp-ibm (dismissed):

woodsp-ibm (approved):

PR #166, opened 2021-06-16
Separate `from_ising` and `to_ising` out from `QuadraticProgram`

t-imamichi:
### Summary

Separated `from_ising` and `to_ising` out from `QuadraticProgram` by adding them as translators.

### DIFF
--- a/qiskit_optimization/problems/quadratic_program.py
+++ b/qiskit_optimization/problems/quadratic_program.py
@@ -27,7 +27,6 @@ from qiskit.exceptions import MissingOptionalLibraryError
 from qiskit.opflow import OperatorBase, PauliSumOp
 from scipy.sparse import spmatrix
 
-from ..deprecation import DeprecatedType, deprecate_method
 from ..exceptions import QiskitOptimizationError
 from ..infinity import INFINITY
 from .constraint import Constraint, ConstraintSense
@@ -37,6 +36,7 @@ from .quadratic_constraint import QuadraticConstraint
 from .quadratic_expression import QuadraticExpression
 from .quadratic_objective import QuadraticObjective
 from .variable import Variable, VarType
+from ..deprecation import DeprecatedType, deprecate_method
 
 logger = logging.getLogger(__name__)
 
@@ -1013,6 +1013,7 @@ class QuadraticProgram:
             QiskitOptimizationError: If a variable type is not binary.
             QiskitOptimizationError: If constraints exist in the problem.
         """
+        # pylint: disable=cyclic-import
         from ..translators.ising import to_ising
 
         return to_ising(self)
@@ -1042,6 +1043,7 @@ class QuadraticProgram:
             QiskitOptimizationError: If there are more than 2 Pauli Zs in any Pauli term
             NotImplementedError: If the input operator is a ListOp
         """
+        # pylint: disable=cyclic-import
         from ..translators.ising import from_ising
 
         other = from_ising(qubit_op, offset, linear)
--- a/qiskit_optimization/problems/quadratic_program.py
+++ b/qiskit_optimization/problems/quadratic_program.py
@@ -872,6 +872,7 @@ class QuadraticProgram:
         Raises:
             QiskitOptimizationError: if the model contains unsupported elements.
         """
+        # pylint: disable=cyclic-import
         from ..translators.docplex_mp import from_docplex_mp
 
         other = from_docplex_mp(model)
@@ -890,6 +891,7 @@ class QuadraticProgram:
         Raises:
             QiskitOptimizationError: if non-supported elements (should never happen).
         """
+        # pylint: disable=cyclic-import
         from ..translators.docplex_mp import to_docplex_mp
 
         return to_docplex_mp(self)
@@ -900,6 +902,7 @@ class QuadraticProgram:
         Returns:
             A string representing the quadratic program.
         """
+        # pylint: disable=cyclic-import
         from ..translators.docplex_mp import to_docplex_mp
 
         return to_docplex_mp(self).export_as_lp_string()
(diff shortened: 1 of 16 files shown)


stefan-woerner (commented):
lgtm, let's discuss the type hints in the call later today

t-imamichi:
Thanks to #170, I simplified this PR to remove `TYPE_CHECKING`.

qiskit_optimization/translators/ising.py
```diff
@@ -0,0 +1,255 @@
...
+        offset: The constant value in the Ising Hamiltonian.
+
+    Raises:
+        QiskitOptimizationError: If a variable type is not binary.
+        QiskitOptimizationError: If constraints exist in the problem.
+    """
+    # if problem has variables that are not binary, raise an error
+    if quad_prog.get_num_vars() > quad_prog.get_num_binary_vars():
+        raise QiskitOptimizationError(
+            "The type of variable must be a binary variable. "
```
woodsp-ibm, line 43: How above `The type of all variables must be binary. ` so its plural as otherwise it seems a little odd

qiskit_optimization/translators/ising.py
```diff
@@ -0,0 +1,255 @@
...
+            "Use a QuadraticProgramToQubo converter to convert "
+            "integer variables to binary variables. "
+            "If the problem contains continuous variables, "
+            "currently we can not apply VQE/QAOA directly. "
+            "you might want to use an ADMM optimizer "
+            "for the problem. "
```
woodsp-ibm, line 49: Should we rephrase this in some way. Basically I think its saying that if there are continuous variable there is no path to ising (unlike with integer vars where a convertor can first be used). Then it giving other options for solving the problem i.e. ADMM, without going through the conversion. Which I guess is fine

qiskit_optimization/translators/ising.py
```diff
@@ -0,0 +1,255 @@
...
+            since :math:`x^2 = x` for :math:`x \in \{0,1\}`.
+            Else, :math:`x^2` is treat as a quadratic term.
+            The default value is False.
+
+    Returns:
+        The quadratic program corresponding to the qubit operator.
+
+    Raises:
+        QiskitOptimizationError: If there are Pauli Xs in any Pauli term
+        QiskitOptimizationError: If there are more than 2 Pauli Zs in any Pauli term
```
woodsp-ibm, line 154: Here its using capital X and Z. In the code its checking Pauli.x and Pauli.z the .x will capture any X or Ys. I am just wondering about the text here - should it say 'If there are Pauli Xs or Ys in any term' which seems more correct to me

qiskit_optimization/translators/ising.py
```diff
@@ -0,0 +1,255 @@
...
+        num_z = len(z_index)
+
+        # Add its weight of the Pauli term to the corresponding element of QUBO matrix
+        if num_z == 1:
+            qubo_matrix[z_index[0], z_index[0]] = coeff.real
+        elif num_z == 2:
+            qubo_matrix[z_index[0], z_index[1]] = coeff.real
+        else:
+            raise QiskitOptimizationError(
+                "There are more than 2 Pauli Zs in the Pauli term {}".format(pauli.z)
```
woodsp-ibm, line 201: Again a place where we could use f String instead of format

qiskit_optimization/problems/quadratic_program.py
```diff
@@ -1015,94 +1013,10 @@ def to_ising(self) -> Tuple[OperatorBase, float]:
...
-        # appropriate size
-        if isinstance(qubit_op, OperatorBase):
-            qubit_op = qubit_op.reduce()
-        else:
-            qubit_op = I ^ num_nodes
+        # pylint: disable=cyclic-import
+        from ..translators.ising import to_ising
 
-        return qubit_op, offset
+        return to_ising(self)
```
woodsp-ibm, line 1019: So unlike the to/from docplex methods, that were deprecated here, we are chosing to keep to/from ising methods in the QP and not deprecate them?

qiskit_optimization/translators/ising.py
t-imamichi: Thanks. I replace the message.

qiskit_optimization/translators/ising.py
t-imamichi: I agree. I put your expression "If there are Pauli Xs or Ys in any term".

qiskit_optimization/problems/quadratic_program.py
t-imamichi: I would like to keep these QP methods because they are fundamental features.

woodsp-ibm (dismissed):

t-imamichi:
I notice that we need to raise an error in `from_ising` if any Pauli term has an imaginary coefficients. So, I added an error check and a unit test. In the previous implementation, coefficients were just converted to real values.

qiskit_optimization/translators/ising.py
```diff
@@ -166,90 +153,69 @@ def from_ising(
...
     for pauli_op in pauli_list:
         pauli_op = pauli_op.to_pauli_op()
         pauli = pauli_op.primitive
         coeff = pauli_op.coeff
-        # Count the number of Pauli Zs in a Pauli term
-        lst_z = pauli.z.tolist()
-        z_index = [i for i, z in enumerate(lst_z) if z is True]
+
+        if coeff.imag != 0.0:
+            raise QiskitOptimizationError(f"Imaginary coefficient exists: {pauli_op}")
```
t-imamichi, line 175: I added a error check with imaginary coefficients.

woodsp-ibm: I have seen situations where sometimes we have aspects that are e-17 or very small. If it has a tiny imaginary part due some some computation rounding should we just ignore it. e.g `if abs(coeff.imag) > 1e-15` or whatever suitable threshold only then raise an error. What do you think?

t-imamichi: Good point. I use `math.isclose` for zero check. fc9afe56d97a5a0ef3428c

woodsp-ibm (approved):

PR #167, opened 2021-06-16
QAOA runtime program

### LINKED ISSUE #168: QAOA Runtime Program

Allow executing QAOA in the cloud using Qiskit Runtime.

Cryoris:
### Summary

Add the `VQEProgram` and `QAOAProgram` as Qiskit Runtime programs for Qiskit Optimization.

Closes #168

### Details and comments

This first version allows running standard QAOA or VQE using Qiskit Runtime. In the future we'll add optimization and QAOA specific features, such as using CVaR-expectation or warm-starting QAOA.

Todo: Run some manual tests using the actual runtime.


woodsp-ibm:
I think we need to add 'runtime' to root init.py file in the list of folders there and a stub for qiskit_optimzation.runtime.rst to docs\apidocs

Is there going to be any runtime tutorial added for either of these or do we rely on the one in Nature.

Should we have some more text in the docstring here that talks about these and their relation to optimization and perhaps the one in Nature. If an average user comes here and has seen the runtime in Nature does it answer any questions they might have over which to use and why I see what seems to be the same thing - VQEProgram in both

qiskit_optimization/runtime/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Qiskit Optimization Runtime (:mod:`qiskit_optimization.runtime`)
+================================================================
+
+.. currentmodule:: qiskit_nature.runtime
```
t-imamichi, line 17:
```suggestion
.. currentmodule:: qiskit_optimization.runtime
```

test/runtime/test_qaoaprogram.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Test the VQE program."""
```
1ucian0, line 13:
```suggestion
"""Test the QAOA program."""
```

test/runtime/test_qaoaprogram.py
```diff
@@ -0,0 +1,56 @@
...
+from qiskit.algorithms import VQEResult
+from qiskit.opflow import I, Z
+
+from qiskit_optimization.runtime import QAOAProgram
+
+from .fake_vqeruntime import FakeRuntimeProvider
+
+
+class TestQAOAProgram(QiskitOptimizationTestCase):
+    """Test the VQE program."""
```
1ucian0, line 29:
```suggestion
    """Test the QAOA program."""
```

test/runtime/test_qaoaprogram.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def setUp(self):
+        super().setUp()
+        self.provider = FakeRuntimeProvider()
+
+    def test_standard_case(self):
+        """Test a standard use case."""
+        operator = Z ^ I ^ Z
+        reps = 2
+        initial_point = np.random.random(2 * reps)
```
1ucian0, line 39: Stable tests are good tests :)

```suggestion
        initial_point = np.random.RandomState(42).random(2 * reps)
```

test/runtime/test_vqeprogram.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def setUp(self):
+        super().setUp()
+        self.provider = FakeRuntimeProvider()
+
+    def test_standard_case(self):
+        """Test a standard use case."""
+        circuit = RealAmplitudes(3)
+        operator = Z ^ I ^ Z
+        initial_point = np.random.random(circuit.num_parameters)
```
1ucian0, line 40:
```suggestion
        initial_point = np.random.RandomState(42).random(circuit.num_parameters)
```

qiskit_optimization/runtime/vqe_program.py
```diff
@@ -0,0 +1,374 @@
...
+    """The VQEProgram result object.
+
+    This result objects contains the same as the VQEResult and additionally the history
+    of the optimizer, containing information such as the function and parameter values per step.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._job_id = None  # type: str
+        self._optimizer_history = None  # type: Dict[str, Any]
```
t-imamichi, line 305:
```suggestion
        self._job_id: Optional[str] = None
        self._optimizer_history: Optional[Dict[str, Any]] = None 
```

qiskit_optimization/runtime/qaoa_program.py
```diff
@@ -0,0 +1,159 @@
...
+        self._mixer = mixer
+        self._reps = reps
+
+    @property
+    def ansatz(self) -> Optional[QuantumCircuit]:
+        return self._ansatz
+
+    @ansatz.setter
+    def ansatz(self, ansatz: QuantumCircuit) -> None:
+        raise NotImplementedError(
```
1ucian0, line 95: Is this the right exception to raise in this case?

Cryoris: i replaced it with a `QiskitOptimizationError` since I didn't find any other satisfying replacement 🙂

Cryoris:
> I think we need to add 'runtime' to root init.py file in the list of folders there and a stub for qiskit_optimzation.runtime.rst to docs\apidocs

Added 👍🏻 

> Is there going to be any runtime tutorial added for either of these or do we rely on the one in Nature.

We should add one here as nature does not cover QAOA. Let's do it in a follow-up though I feel like this PR is large enough as it is.

> Should we have some more text in the docstring here that talks about these and their relation to optimization and perhaps the one in Nature. If an average user comes here and has seen the runtime in Nature does it answer any questions they might have over which to use and why I see what seems to be the same thing - VQEProgram in both - we only have one VQE in Terra after all that is used by either but not VQEProgram

Well right now the VQEPrograms are 1:1 the same thing, and they might stay the same! But since we decided not to put it in Terra, having the copy decouples Optimization and Nature. I'll put a note that they are the same.

test/runtime/test_vqeprogram.py
```diff
@@ -0,0 +1,57 @@
...
+        vqe = VQEProgram(
+            ansatz=circuit,
+            optimizer=optimizer,
+            initial_point=initial_point,
+            backend=backend,
+            provider=self.provider,
+        )
+        result = vqe.compute_minimum_eigenvalue(operator)
+
+        self.assertIsInstance(result, VQEResult)
```
woodsp-ibm, line 53: I guess a VQEProgramResult is a VQEResult since that is its parent. But what do we expect here - a VQEProgramResult - should we rather not be testing for that instead - or additionally

test/runtime/test_qaoaprogram.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_standard_case(self):
+        """Test a standard use case."""
+        operator = Z ^ I ^ Z
+        reps = 2
+        initial_point = np.random.RandomState(42).random(2 * reps)
+        optimizer = {"name": "SPSA", "maxiter": 100}
+        backend = QasmSimulatorPy()
+
+        vqe = QAOAProgram(
```
woodsp-ibm, line 43: Very minor comment - it might make more sense that the var name was `qaoa` rather than `vqe` (presumably a left over from copy/paste) but in the end it does not really matter too much.

test/runtime/test_qaoaprogram.py
```diff
@@ -0,0 +1,56 @@
...
+            optimizer=optimizer,
+            reps=reps,
+            initial_point=initial_point,
+            backend=backend,
+            provider=self.provider,
+        )
+        result = vqe.compute_minimum_eigenvalue(operator)
+
+        self.assertIsInstance(result, VQEResult)
```
woodsp-ibm, line 52: Similar comment about the type check

qiskit_optimization/runtime/vqe_program.py
```diff
@@ -0,0 +1,378 @@
...
+        vqe_result.cost_function_evals = result.get("cost_function_evals", None)
+        vqe_result.eigenstate = result.get("eigenstate", None)
+        vqe_result.eigenvalue = result.get("eigenvalue", None)
+        vqe_result.aux_operator_eigenvalues = result.get("aux_operator_eigenvalues", None)
+        vqe_result.optimal_parameters = result.get("optimal_parameters", None)
+        vqe_result.optimal_point = result.get("optimal_point", None)
+        vqe_result.optimal_value = result.get("optimal_value", None)
+        vqe_result.optimizer_evals = result.get("optimizer_evals", None)
+        vqe_result.optimizer_time = result.get("optimizer_time", None)
+        vqe_result.optimizer_history = result.get("optimizer_history", None)
```
dlasecki, line 294: Maybe extract it to a separate method?

qiskit_optimization/runtime/vqe_program.py
```diff
@@ -0,0 +1,378 @@
...
+
+        # define runtime options
+        options = {"backend_name": self.backend.name()}
+
+        # send job to runtime and return result
+        job = self.provider.runtime.run(
+            program_id=self.program_id,
+            inputs=inputs,
+            options=options,
+            callback=self._wrap_vqe_callback(),
```
dlasecki, line 274: Maybe extract it to a separate method?

qiskit_optimization/runtime/vqe_program.py
```diff
@@ -0,0 +1,378 @@
...
+
+def _convert_to_paulisumop(operator):
+    """Attempt to convert the operator to a PauliSumOp."""
+    if isinstance(operator, PauliSumOp):
+        return operator
+
+    try:
+        primitive = SparsePauliOp(operator.primitive)
+        return PauliSumOp(primitive, operator.coeff)
+    except Exception as exc:
```
dlasecki, line 374: Can we catch any more specific exception here?

test/runtime/fake_vqeruntime.py
```diff
@@ -0,0 +1,99 @@
...
+
+from typing import Dict, Any
+import numpy as np
+from qiskit.circuit import QuantumCircuit
+from qiskit.opflow import PauliSumOp
+from qiskit.providers import Provider
+from qiskit_optimization.runtime import VQEProgramResult
+
+
+class FakeVQEJob:
```
dlasecki, line 23: Would using the mock library be a better practice here? 
https://docs.python.org/3/library/unittest.mock.html

dlasecki (changes requested):

qiskit_optimization/runtime/vqe_program.py
Cryoris: I'm not sure what exceptions should come up here, so I just catched any and gave more context. The alternative would be to not try-except at all and just let the code fail without an extra message

test/runtime/fake_vqeruntime.py
Cryoris: Maybe, I'm not familiar with using the mock library yet and it seemed fairly straightforward to just implement these classes. Would it be easier or shorter with the mock library?

qiskit_optimization/runtime/vqe_program.py
woodsp-ibm: As this code is a duplicate of the one in Nature, from my viewpoint its best to continue to keep the code identical, so that hopefully in the future, it can go in one place and then we don't have to reconcile differences between them. Hence if this is done here I think it should be done in Nature. Since we are nearing a release with everyone busy my reaction would be to let it be as-is at present.

test/runtime/test_vqeprogram.py
Cryoris: yes, good point, I'll update

qiskit_optimization/runtime/vqe_program.py
Cryoris: I already changed this locally, I don't mind either way. Since it's already done I'll push it now but we can also revert.

dlasecki: We might keep the version above for the sake of consistency as Steve suggests.

qiskit_optimization/runtime/vqe_program.py
dlasecki: I see. We may keep it. But when it is obvious which kind of exceptions we expect, it is best to specify them and handle accordingly. In the above, we may have some wild exception and it would actually inform the user that the type of the operator is invalid, which may not be the problem...

woodsp-ibm:
There is still a todo in the main description above.

>Todo: Run some manual tests using the actual runtime.

Has this been done that we are happy with everything here now?

qiskit_optimization/runtime/vqe_program.py
Cryoris: Yeah that's why I'm re-raising the original one, so that the information is not lost 🙂

dlasecki (approved):

t-imamichi (approved):

Cryoris:
@woodsp-ibm yes we do 👍🏻

PR #157, opened 2021-06-08
Special patterns of inequality constraint to penalty converters

knamba-jp:
### Summary
Resolves #40. For details about previous discussion, see #98 

### Details and comments
Owned by Amano-san and me

[A Tutorial on Formulating and Using QUBO Models](https://arxiv.org/abs/1811.11538) introduces examples of the special patterns in page 10 as follows.
image

PR #169, opened 2021-06-16
Add lower bound and upper bound of `LinearExpression` and `QuadraticExpression`

t-imamichi:
### Summary

Add `LinearExpression.bounds()` and `QuadraticExpression.bounds()` to obtain a tuple of the lower bound and the upper bound.
They can be used for `LinearEqualityToPenalty`, `LinearInequalityToPenalty`, and `from_docplex_mp` (indicator constraint) and this PR may simplifies them.

This also simplifies `LinearEqualityToPenalty` and `InequalityToEquality` using the bounds.
This can be used for #157 too. But it does not work for #163.

TODO:
- [x] Discussion whether we need this feature or not.


qiskit_optimization/problems/linear_expression.py
```diff
@@ -172,3 +173,41 @@ def evaluate_gradient(self, x: Union[ndarray, List, Dict[Union[int, str], float]
 
         # extract the coefficients as array and return it
         return self.to_array()
+
+    def _bounds(self) -> Tuple[float, float]:
```
woodsp-ibm, line 177: This could easily be public method and made a property so you can just do `l_b,  u_b = le.bounds`.

In place of have other methods/properties you could use a `namedtuple` so you could access the values by sat 'lower' and 'upper' e.g `le.bounds.lower` (the namedtuple just makes it a little more readable than doing `le.bounds[0]`

As to the larger question of whether this is needed or not I have no insight to that to offer.

t-imamichi:
I notice that `from_docplex_mp` (indicator constraint) cannot use this feature to compute LB and UB. `LinearEqualityToPenalty` does not always require this because it handles a easier case (all variables are converted into binary).
This PR is necessary only for `LinearInequalityToPenalty` #157. So, we don't need this PR and #157 can implement the same feature as part of their converter.

t-imamichi:
I found this can be used in `InequalityToEquality` as well as `LinearInequalityToPenalty` and `EqualityToPenalty`.
So, I want to revise this PR to simplify the converters.
https://github.com/Qiskit/qiskit-optimization/blob/cfef1fee1185b82222a11fce846885e655c8fce5/qiskit_optimization/converters/inequality_to_equality.py#L275
https://github.com/Qiskit/qiskit-optimization/blob/cfef1fee1185b82222a11fce846885e655c8fce5/qiskit_optimization/converters/inequality_to_equality.py#L304
https://github.com/Qiskit/qiskit-optimization/blob/main/qiskit_optimization/converters/linear_equality_to_penalty.py#L159-L162

qiskit_optimization/problems/linear_expression.py
t-imamichi: Thank you for your idea. Namedtuple or dataclass sounds better.

qiskit_optimization/problems/linear_expression.py
```diff
@@ -172,3 +185,25 @@ def evaluate_gradient(self, x: Union[ndarray, List, Dict[Union[int, str], float]
 
         # extract the coefficients as array and return it
         return self.to_array()
+
+    def bounds(self) -> ExpressionBounds:
```
woodsp-ibm, line 189: Would this be easier for callers as a property? Instead of doing `problem.objective.quadratic.bounds().upperbound` where `bounds()` needs to be done like that, it could instead be `problem.objective.quadratic.bounds.upperbound`

qiskit_optimization/problems/quadratic_expression.py
```diff
@@ -241,3 +243,44 @@ def _cast_as_array(
         if isinstance(x, list):
             x = np.array(x)
         return x
+
+    def bounds(self) -> ExpressionBounds:
```
woodsp-ibm, line 247: Similar comment to the linear expression above

qiskit_optimization/problems/linear_expression.py
t-imamichi: Thanks. Property looks smart. I made it property.

test/converters/test_converters.py
```diff
@@ -110,10 +110,10 @@ def test_inequality_binary(self):
                 "x0",
                 "x1",
                 "x2",
-                "x1x2@int_slack",
-                "x0x2@int_slack",
-                "x0x1_x1x2LE@int_slack",
-                "x0x1_x1x2GE@int_slack",
+                "x1x2@integer_slack",
```
woodsp-ibm, line 113: This is the public `name` of the variables we create during the conversion right. Hopefully no user is accessing these variables and using the names, in which case, like the test here things would fail for them. Since this is really an internal convertor issue - and I guess the names really are just a choice to have something meaningful/readable - it seems fine to me. If anyone is using them, which I think is unlikely, then its easy enough for them to change as was done here.

woodsp-ibm (approved):
Seems fine to me. I made an observation about the naming but I think its ok as-is.

test/converters/test_converters.py
t-imamichi: Thank you for your feedback. We don't have to take the risk to change the public name now. I reverted it.

woodsp-ibm (approved):

a-matsuo (approved):

PR #170, opened 2021-06-16
Remove TYPE_CHECKING flag from code

manoelmarques:
### Summary
Remove TYPE_CHECKING and add QuadraticProgram to typehints.
Add translators to docs


t-imamichi (approved):
Thanks. It's a much simpler solution.

PR #174, opened 2021-06-22
Refactor converters and `MinimumEigenOptimizer`

t-imamichi:
### Summary

- Changed `MinimumEigenOptimizer` to return the best solution in terms of the original problem. See details.
- Simplified `MinimumEigenOptimizer` and `QuadraticProgramToQubo`
- Removed unnecessary deepcopy in converters
- Revised type hints
- Replace error messages with f-strings

Note: need to address  #175 to pass CI

### Details and comments

The current implementation `MinimumEigenOptimizer` returns the best solution `raw_samples[0].x` among `raw_samples` in terms of QUBO as `MinimumEigenOptimizationResult.x`. But, it is not always the best solution in terms of the original problem because inequality constraints are converted into penalties as part of QUBO. So, I changed `MinimumEigenOptimizer` to return the best solution in terms of the original problem as `MinimumEigenOptimizationResult.x`.

### DIFF
--- a/qiskit_optimization/algorithms/optimization_algorithm.py
+++ b/qiskit_optimization/algorithms/optimization_algorithm.py
@@ -15,7 +15,7 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from enum import Enum
-from typing import List, Union, Any, Optional, Dict, Type, cast
+from typing import List, Union, Any, Optional, Dict, Type, Tuple, cast
 from warnings import warn
 
 import numpy as np
@@ -99,7 +99,7 @@ class OptimizationResult:
     def __init__(
         self,
         x: Optional[Union[List[float], np.ndarray]],
-        fval: float,
+        fval: Optional[float],
         variables: List[Variable],
         status: OptimizationResultStatus,
         raw_results: Optional[Any] = None,
@@ -126,10 +126,8 @@ class OptimizationResult:
         else:
             if len(x) != len(variables):
                 raise QiskitOptimizationError(
-                    "Inconsistent size of optimal value and variables. x: size {} {}, "
-                    "variables: size {} {}".format(
-                        len(x), x, len(variables), [v.name for v in variables]
-                    )
+                    f"Inconsistent size of optimal value and variables. x: size {len(x)} {x}, "
+                    f"variables: size {len(variables)} {[v.name for v in variables]}"
                 )
             self._x = np.asarray(x)
             self._variables_dict = dict(zip(self._variable_names, self._x))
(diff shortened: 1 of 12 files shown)


qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -99,7 +99,7 @@ class OptimizationResult:
     def __init__(
         self,
         x: Optional[Union[List[float], np.ndarray]],
-        fval: float,
+        fval: Optional[float],
```
woodsp-ibm, line 102: If fval is passed in as None I imagine that is how it remains - so the fval getter now should be updated for its return typehint o say Optional[float] too right?

t-imamichi: If `x` is `None`, `fval` is needed to be `None` as well. It happens in this line. https://github.com/Qiskit/qiskit-optimization/blob/7f341b00d850577cf866caecbee61a102144156e/qiskit_optimization/algorithms/minimum_eigen_optimizer.py#L235-L238
But, it perhaps won't happen according to https://github.com/Qiskit/qiskit-optimization/blob/7f341b00d850577cf866caecbee61a102144156e/qiskit_optimization/algorithms/optimization_algorithm.py#L501
Let me check more details.

t-imamichi: `eigen_result.eigenstate` can be `None` because eigenstate is `Optional[ndarray]`. So, `fval` needs to be `None` in case `eigenstate` is `None`.
https://github.com/Qiskit/qiskit-optimization/blob/7f341b00d850577cf866caecbee61a102144156e/qiskit_optimization/algorithms/minimum_eigen_optimizer.py#L218
```python
class MinimumEigensolverResult(AlgorithmResult):
    """Minimum Eigensolver Result."""
...
    @property
    def eigenstate(self) -> Optional[np.ndarray]:
        """return eigen state"""
        return self._eigenstate
```
https://github.com/Qiskit/qiskit-terra/blob/0167c156d2e2c929fd9041ef75762905ae826ccc/qiskit/algorithms/minimum_eigen_solvers/minimum_eigen_solver.py#L87

t-imamichi: I updated the type of `OptimizationResult.fval` to be `Optional[float]`.

t-imamichi:
Due to #175, CI fails.

t-imamichi:
#176 fixed #175

woodsp-ibm:
Should we have a reno for this, due to the change in behavior, since it may return different values for an end user than it did in the past.

> So, I changed MinimumEigenOptimizer to return the best solution in terms of the original problem as MinimumEigenOptimizationResult.x.

t-imamichi:
Yes. I added a reno.

woodsp-ibm (approved):
LGTM

PR #176, opened 2021-06-23
Add type ignore for numpy 1.21

### LINKED ISSUE #175: Mypy reports an error with Numpy 1.21
### Information

- **Python version**: python 3.7.10, python 3.8.5
- **Operating system**: ubuntu 20.04.2

`make mypy` reports the following error with the main branch.
```
mypy qiskit_optimization test tools
qiskit_optimization/algorithms/admm_optimizer.py:685: error: Incompatible types in assignment (expression has type "ndarray[Any, Any]", variable has type "QuadraticExpression")
Found 1 error in 1 file (checked 99 source files)
```

This happens with numpy 1.21 (released 3 hours ago). https://github.com/numpy/numpy/releases/tag/v1.21.0
Mypy does not report the error with numpy 1.20.

The following line causes the error. Mypy cannot perhaps handle setter `quadratic` properly.
https://github.com/Qiskit/qiskit-optimization/blob/7f341b00d850577cf866caecbee61a102144156e/qiskit_optimization/algorithms/admm_optimizer.py#L685
Interestingly, a similar code (setter of `linear`) is treated as no error. I cannot see any difference between them.
https://github.com/Qiskit/qiskit-optimization/blob/7f341b00d850577cf866caecbee61a102144156e/qiskit_optimization/algorithms/admm_optimizer.py#L689

This may be related to https://github.com/python/mypy/issues/3004

`make mypy` with numpy 1.21.

no error

### Suggested solutions

Shall we add `# type: ignore`?

t-imamichi:
### Summary

Fixes #175 

I looked for a solution of the mypy error, but I could not find one.
I welcome better solutions.

### DIFF
--- a/qiskit_optimization/algorithms/admm_optimizer.py
+++ b/qiskit_optimization/algorithms/admm_optimizer.py
@@ -682,7 +682,7 @@ class ADMMOptimizer(OptimizationAlgorithm):
         quadratic_y = self._params.beta / 2 * np.eye(binary_size) + self._state.rho / 2 * np.eye(
             binary_size
         )
-        op3.objective.quadratic = quadratic_y
+        op3.objective.quadratic = quadratic_y  # type: ignore[assignment]
 
         # set linear objective for y
         linear_y = -self._state.lambda_mult - self._state.rho * (self._state.x0 - self._state.z)


manoelmarques:
Maybe a cast would work but this is holding PRs, if there is a better solution it can be fixed in another PR

manoelmarques (approved):

t-imamichi:
A cast `cast(QuadraticExpression, quadratic)` resolves the error. But, it is a wrong cast because `quadratic` is not `QuadraticExpression`. So I chose `type: ignore[assignment]` instead

PR #178, opened 2021-06-24
Suppress Gurobi message when `disp==False` and add the missing model update

t-imamichi:
### Summary

- Removed unnecessary `gurobipy.Model.optimize()` in test_gurobi.py
- Added a missing model update in `to_gurobipy`
- Suppress the following message of parameter change in `GurobiOptimizer` if `disp` is False. 
```
Changed value of parameter NonConvex to 2
   Prev: -1  Min: -1  Max: 2  Default: -1
```

### DIFF
--- a/qiskit_optimization/algorithms/gurobi_optimizer.py
+++ b/qiskit_optimization/algorithms/gurobi_optimizer.py
@@ -124,13 +124,12 @@ class GurobiOptimizer(OptimizationAlgorithm):
         # convert to Gurobi problem
         model = to_gurobipy(problem)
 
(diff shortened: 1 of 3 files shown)


woodsp-ibm (approved):
When you say suppresses it - it seems more like you moved the statement beyond the self.disp  test, that seems to control output, so that the message, rather than being always there, is now dependent, as would be expected, on the `disp` flag.

t-imamichi:
Sorry. More precisely, that parameter message is displayed even if `disp==False`. This PR suppress it when `disp==True`.

Issue #77, opened 2021-04-13
Add textual representations to optimization objects


adekusar-drl:
### What is the expected enhancement?
Many optimization objects, e.g. `Constraint`, `Variable` do not have textual representations, e.g. methods like `__str__` and `__repr__` and in a debugger, e.g. PyCharm they don't provide much information out of box. `QuadraticProgram` has only `__repr__` which, in turns, depends on the docplex implementation, and in the debugger you can only see a docplex comment header of the LP representation.

This issue should be fixed by reviewing optimization objects and introducing textual representation methods in them.

t-imamichi:
The current `__repr__` does not follow the Python's convention. We should fix it.

> `object.__repr__(self)`
> Called by the repr() built-in function to compute the “official” string representation of an object. If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.
https://docs.python.org/3/reference/datamodel.html#object.__repr__

> `object.__str__(self)`
> Called by str(object) and the built-in functions format() and print() to compute the “informal” or nicely printable string representation of an object. The return value must be a string object.
https://docs.python.org/3/reference/datamodel.html#object.__str__

PR #183, opened 2021-06-29
Add a tutorial of translators and classical optimization solvers

t-imamichi:
### Summary

Adds a tutorial about translators and classical optimizers (CPLEX and Gurobi).


a-matsuo:
Shouldn't this be called as a tutorial for `translator` instead of summary of 2.0?

t-imamichi:
Good point. This includes both translators and GurobiOptimizer, so I named it a summary of 0.2.0.
But, I can separate the translator part as an independent tutorial section. I'm wondering which is better.

woodsp-ibm:
Bear in mind we have the release notes that show up in the documentation that include text etc around new features that have been added. I am sure some features might warrant more explanation etc in a tutorial, whether new tutorial(s) or updating existing one(s) - maybe its best to have the tutorials around a learning path for the code etc and not mix purpose with the release notes.

t-imamichi:
Thank you for your comment. I agree with you. I will split this notebook to write new tutorial(s) or update existing one(s).

t-imamichi:
I updated it as a new tutorial about translators and classical optimizers.

woodsp-ibm:
In the title it refers to `classical optimizers` and in the text there is also `classical solvers`. Maybe the title could be something like `Working with classical optimization solvers` or `Using classical optimization solvers with Qiskit` then translators that take a model from that, and algorithms here directly using it e.g. Cplex & Gurboi optimizers...

I am seeing Cplex and Gurobi mentioned as two of these solvers upfront yet the tutorial starts off with Docplex/Gurobi. So we now have what seems like three things - Cplex & Docplex and Gurobi. Perhaps the relationship between Docplex & Cplex could be clarified. I guess we have a docplex & gurobi convertors (not cplex & gurobi) but cplex and gurobi optimizers. Maybe if you are coming here for the first time this is a bit confusing!

Minor comment - we should add the standard Qiskit version widget at the end

t-imamichi:
Thank you. I clarified the solver (CPLEX and Gurobi) and Python API (Docplex and Gurobipy), updated the title as `Working with classical optimization solvers`, and added the version widget.

docs/tutorials/11_working_with_classical_optimization_solvers.ipynb
```diff
@@ -0,0 +1,469 @@
...
+   "cell_type": "markdown",
+   "metadata": {}
+  },
+  {
+   "source": [
+    "We can use classical optimization solvers (CPLEX and Gurobi) with Qiskit Optimization.\n",
+    "Docplex and Gurobipy are the Python APIs for CPLEX and Gurobi, respectively.\n",
+    "We can load and save an optimization model by Docplex and Gurobipy and can apply CPLEX and Gurobi to `QuadraticProgram`.\n",
+    "\n",
+    "If you want to use the solver of CPLEX, you need to install `pip install 'qiskit-optimization[cplex]'`. Docplex is supposed to be installed with Qiskit Optimization.\n",
```
woodsp-ibm, line 40: Would this `If you want to use the solver of CPLEX...` be better as `If you want to use the CPLEX solver`

woodsp-ibm:
In the notebook bullet points that start on the next line right after paragraph text come out as a bullet list. Sphinx however needs to have a blank line preceding the list otherwise it considers the text part of the same paragraph and wraps it in. E.g here where in the notebook the links are two bullet points - here they have been included are part of the same paragraph

image

woodsp-ibm:
I guess this in the output of cell [3], when it uses export as LP string to print the gurobi model might be confusing to users?
>from gurobipy
\ ENCODING=ISO-8859-1
\Problem name: gurobi
>
> Maximize
 obj: [ 2 x*y ]/2
Subject To
 R0: x - y <= 0
 
Mainly because it also shows says generated by DOCplex - now that function is of course in DOCplex so if you know that then maybe its less confusing to see DOCplex intermingled with the Gurobi model! Maybe this new str printout that is being worked on, if/when thats used instead, would not be confusing in this way.

woodsp-ibm:
In Cell [5] where it talks about indicator contraints I guess the first part, before it says ` solves the problem that has the indicator constraint. I assume this line `print(ind_mod.solve())` - what is being used to solve the problem here - DOCplex itself? or does it in fact use CPLEX to solve it there direct from the model? Often we show some Qiskit classical solver and compare to that - in this case I has to look back and see what was happening after I read the text `We get the same solution by QAOA` to see that not only was it showing and indicator contraint and how it converted to the model, but also solving the problem. It might be clearer to put the solve into its own cell after printing the model. That way you have the model, a classical solution and the same outcome with QAOA in three steps/cells.

docs/tutorials/11_working_with_classical_optimization_solvers.ipynb
t-imamichi: Thanks. I updated it.

t-imamichi:
As for Cell [3], I agree that it is a bit confusing to have a header "Generated by Docplex". When #77 is merged, I will update this tutorial.

As for Cell [5], because CPLEX solver supports the indicator constraints natively, I want to compare the solutions with and without `from_docplex_mp` to say that the translator does not change the optimal solution. I add some more description to avoid the confusion.

I also reorganized the sections to put classical solvers first then translators because the classical solver section is shorter.

docs/tutorials/11_working_with_classical_optimization_solvers.ipynb
```diff
@@ -0,0 +1,578 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can use classical optimization solvers (CPLEX and Gurobi) with Qiskit Optimization.\n",
+    "Docplex and Gurobipy are the Python APIs for CPLEX and Gurobi, respectively.\n",
+    "We can load and save an optimization model by Docplex and Gurobipy and can apply CPLEX and Gurobi to `QuadraticProgram`.\n",
+    "\n",
+    "If you want to use the CPLEX solver, you need to install `pip install 'qiskit-optimization[cplex]'`. Docplex is supposed to be installed with Qiskit Optimization.\n",
```
woodsp-ibm, line 18: Instead of

> Docplex is supposed to be installed with Qiskit Optimization.

how about

> Docplex is automatically installed, as a dependent, when you install Qiskit Optimization.

docs/tutorials/11_working_with_classical_optimization_solvers.ipynb
```diff
@@ -0,0 +1,578 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Working with Classical Optimization Solvers"
```
woodsp-ibm, line 7: I was wondering if there was a "better" title that `Working with Classical Optimization Solvers` I was thinking about something like `Using Classical Optimization Solvers and Models with Qiskit` or `Interoperating with Classical Optimization Solvers and Models` - something that perhaps in the title makes it feel more like they are being talked about/used in the context of Qiskit Optimization. I added Model there too since we use the representation and I think can do so independent of the solvers.

t-imamichi: OK. I renamed "Using Classical Optimization Solvers and Models with Qiskit Optimization".

woodsp-ibm (approved):

t-imamichi:
Thanks!

PR #184, opened 2021-06-29
[WIP] refactor `__repr__` and `__str__` of `OptimizationResult` and `QuadraticProgram`

### LINKED ISSUE #77: Add textual representations to optimization objects

t-imamichi:
### Summary

Note:
This PR is almost ready, but it gets larger than I expected.
So I will split this into three PRs to make review easier.

- [x] #366
- [ ] #369
- [ ] adds `__str__` and `__repr__` to `QuantumProgram` and update all tutorials
-----------

Refactor existing `__repr__` and `__str__` to comply with Python's convention.
I implement a simpler format `QuadraticProgram.__str__` than LP format because it does not look intuitive especially quadratic terms (it requires 1/2, e.g, `obj: 2 x + 3 y + [ 2 x*u ]/2`) and variable types.
If users need the exact LP format, they can use `export_as_lp_string()` via docplex.
This is the [spec of LP format](https://www.ibm.com/docs/en/icos/20.1.0?topic=cplex-lp-file-format-algebraic-representation).

> `object.__repr__(self)`
> Called by the repr() built-in function to compute the “official” string representation of an object. If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.
https://docs.python.org/3/reference/datamodel.html#object.__repr__

> `object.__str__(self)`
> Called by str(object) and the built-in functions format() and print() to compute the “informal” or nicely printable string representation of an object. The return value must be a string object.
https://docs.python.org/3/reference/datamodel.html#object.__str__

Fixes #77 

TODO
- [x] Unit test
- [x] Discussion of format
- [x] Check tutorials that use the current `__repr__`.
- [ ] Reno

### Details and comments

Update (July 2, 2021): I updated the format based on the discussion https://github.com/Qiskit/qiskit-optimization/pull/184#issuecomment-872326583

```python
from qiskit_optimization.problems import QuadraticProgram

qp = QuadraticProgram('problem 1')
qp.integer_var(-1, 2, 'x')
qp.integer_var(-1, 2, 'y')
qp.continuous_var(-1, name='z')
qp.binary_var('u')
qp.binary_var_list(10)
qp.integer_var_list(3)
qp.continuous_var_list(3)
qp.minimize(constant=3, linear={'x': 2, 'y': 3}, quadratic={('u', 'x'): -1})
qp.linear_constraint({'x': 1, 'y': -2}, '>=', 2, name='lin_GE')
qp.linear_constraint({'x': 2, 'y': -1}, '==', 1, name='lin_EQ')
qp.quadratic_constraint({'x': 1, 'u': 1}, {(3, 4): 1, (5, 6): -1}, '<=', 1, name='quad_LE')
qp.quadratic_constraint({'x': 2, 'y': -1}, {('z', 'z'): -1}, '<=', 1)
print("repr:", repr(qp))
print()
print(qp)

qp = QuadraticProgram('problem 2')
qp.binary_var('x')
qp.integer_var(-1, 2, 'y')
qp.maximize(quadratic={('x', 'y'): 1})
qp.linear_constraint({'x': 1, 'y': -1}, '<=', 1)
print("repr:", repr(qp))
print()
print(qp)

qp = QuadraticProgram('problem 3')
qp.binary_var('x')
qp.integer_var(-1, 2, 'y')
qp.maximize(quadratic={('x', 'y'): 1})
print("repr:", repr(qp))
print()
print(qp)

qp = QuadraticProgram()
print("repr:", repr(qp))
print()
print(qp)
```

```
repr: <QuadraticProgram at 0x7fb90312e5b0: name='problem 1', obj_sense=minimize, num_variables=20, num_constraints=4>

Problem name: problem 1

Minimize
  -x * u + 2x + 3y + 3

Subject to
  Linear constraints (2)
    x - 2y >= 2  'lin_GE'
    2x - y == 1  'lin_EQ'

  Quadratic constraints (2)
    u * x4 - x5 * x6 + x + u <= 1  'quad_LE'
    -z^2 + 2x - y <= 1  'q1'

  Integer variables (5)
    -1 <= x <= 2
    -1 <= y <= 2
    0 <= x14
    0 <= x15
    0 <= x16

  Continuous variables (4)
    -1 <= z
    0 <= x17
    0 <= x18
    0 <= x19

  Binary variables (11)
    u x4 x5 x6 x7 x8 x9 x10 x11 x12 x13

repr: <QuadraticProgram at 0x7fb8e20c0520: name='problem 2', obj_sense=maximize, num_variables=2, num_constraints=1>

Problem name: problem 2

Maximize
  x * y

Subject to
  Linear constraints (1)
    x - y <= 1  'c0'

  Integer variables (1)
    -1 <= y <= 2

  Binary variables (1)
    x

repr: <QuadraticProgram at 0x7fb8e20c0d30: name='problem 3', obj_sense=maximize, num_variables=2, num_constraints=0>

Problem name: problem 3

Maximize
  x * y

Subject to
  No constraints

  Integer variables (1)
    -1 <= y <= 2

  Binary variables (1)
    x

repr: <QuadraticProgram at 0x7fb8e20c0cd0: name='', obj_sense=minimize, num_variables=0, num_constraints=0>

Problem name: 

Minimize
  0

Subject to
  No constraints

  No variables

```

LP format of "problem 1" is as follows. I think 1/2 of the quadratic terms and variable types are not intuitive. `Generals` means integer variables and neither `Binaries` nor `Generals` means continuous variables.
```
# This PR
Problem name: problem 1

Minimize
  -x * u + 2x + 3y + 3

Subject to
  Linear constraints (2)
    x - 2y >= 2  'lin_GE'
    2x - y == 1  'lin_EQ'

  Quadratic constraints (2)
    u * x4 - x5 * x6 + x + u <= 1  'quad_LE'
    -z^2 + 2x - y <= 1  'q1'

  Integer variables (5)
    -1 <= x <= 2
    -1 <= y <= 2
    0 <= x14
    0 <= x15
    0 <= x16

  Continuous variables (4)
    -1 <= z
    0 <= x17
    0 <= x18
    0 <= x19

  Binary variables (11)
    u x4 x5 x6 x7 x8 x9 x10 x11 x12 x13
```
```
# LP format
\Problem name: problem 1

Minimize
 obj: 2 x + 3 y + [ - 2 x*u ]/2 + 3
Subject To
 lin_GE: x - 2 y >= 2
 lin_EQ: 2 x - y = 1
 quad_LE: [ u*x4 - x5*x6 ] + x + u <= 1
 q1: [ - z^2 ] + 2 x - y <= 1

Bounds
 -1 <= x <= 2
 -1 <= y <= 2
 -1 <= z
 0 <= u <= 1
 0 <= x4 <= 1
 0 <= x5 <= 1
 0 <= x6 <= 1
 0 <= x7 <= 1
 0 <= x8 <= 1
 0 <= x9 <= 1
 0 <= x10 <= 1
 0 <= x11 <= 1
 0 <= x12 <= 1
 0 <= x13 <= 1

Binaries
 u x4 x5 x6 x7 x8 x9 x10 x11 x12 x13

Generals
 x y x14 x15 x16
End
```


woodsp-ibm:
Do we need the extra space around constant multipliers e.g. `x * u + 2 x + 3 y` would it be easier to read as `x * u + 2x  + 3y`?

For binary variables do we need the bounds i.e. `0 <= u <= 1: binary` would simply saying `u: binary` suffice - or maybe we need to be explicit that the binary values are 0 and 1?

In LP format I see it uses [ ] (square brackets around some parts. Is that to highlight quadratic components in the expression? Not knowing the format it does look a little odd to me though - if that's what its for I can see its simpler without as you have.

It will be good to make the repr do what its expected - currently with exceptions when it refers back to QP they print that long LP string which seems out of place.

t-imamichi:
Thank you for your feedback.

1. Yes, I will remove the extra space around multipliers.
2. Yes, I will simplify binary variables by removing the range.
3. Yes, LP format uses [ ] to highlight quadratic terms [spec](https://www.ibm.com/docs/en/icos/20.1.0?topic=representation-quadratic-terms-in-lp-file-format). I think my current format without [ ] looks simpler.
4. Agree. I want to make `repr` short but informative. What do you think of my current format? `<QuadraticProgram: name=test, sense=minimize, num_variables=(bin=1, int=2, cont=1), num_constraints=(lin=2, quad=1)>`

woodsp-ibm:
> Yes, LP format uses [ ] to highlight quadratic terms spec. I think my current format without [ ] looks simpler.
I agree its simpler without if they add no real value

> Agree. I want to make repr short but informative. What do you think of my current format
It looked fine to me.

And in looking more at what you include in the repr - counts - I wonder if that might be helpful in the string too

Here since the items have a numeric suffix in the key, and will be in order, the count is simply the max_index of c or q plus 1 for the respective constraints (I guess q is for quadratic - and I assume l is not used for linear for reasons that lowercase L is not used in variable names - c I guess has some commonly known meaning)
```
Subject to
  c0: x - 2 y >= 2
  c1: 2 x - y == 1
  q0: - x * y + 2 x - y <= 1
```  
However in this section there are no numerics like that to aid counting. While the example below is small, and easy enough to count manually
```
Variables
  -1 <= x <= 2: integer
  -1 <= y <= 2: integer
  0 <= u <= 1: binary
  -1 <= z: continuous
```
I wonder if something like this might be helpful for when problems get bigger and its not so easy
```
Variables: 4 total (2 integer, 1 binary, 1 continuous)
  -1 <= x <= 2: integer
  -1 <= y <= 2: integer
  0 <= u <= 1: binary
  -1 <= z: continuous
 ```  
While something like the above could be done with the constraints too it seemed with the numeric part in the key listing them might easily suffice. Of course one could use numeric indexes here and make it more like
```
Variables
  i0: -1 <= x <= 2
  i1: -1 <= y <= 2
  b0: 0 <= u <= 1
  c0: -1 <= z
```

**Comment by t-imamichi** [comment]
The keys of constraints (e.g., c1, q1) are names of constraints. If users don't define name, we assign the default name "c" (linear) or "q" (quadratic) + number. If users assign the constraint names, they appear as keys. In case of variables, "x", "y" are names. So, I don't think we need to add additional index i1 or b0. 
The constraint format is as follows.
```
(constraints)
  name: lhs sense rhs

(integer or continuous variables)
  lb <= name <= ub: type

(binary variables)
  name: type 
```

I'm removing the range of binary variables. So, it does not look well balanced. I'm thinking of moving variable type as a key.
```
Variables
  -1 <= x <= 2: integer
  -1 <= y <= 2: integer
  u: binary
  -1 <= z: continuous
```

idea
```
Variables
  integer:    -1 <= x <= 2
  integer:    -1 <= y <= 2
  binary:     u
  continuous: -1 <= z:
```

**Comment by woodsp-ibm** [comment]
Ok I did not realize those were the default names of the constraints. Sure your proposal for variables `type: range` seems fine. I think it would be nice though to have the summary on the `Variables` line e.g as I showed above `Variables: 4 total (2 integer, 1 binary, 1 continuous)` for when there are a lot of variables its handy - same for `Subject to` now then as the c0 etc are names it would be nice to say something like `Subject to: 2 linear and 4 quadratic constraints`

**Comment by t-imamichi** [comment]
I agree on the number of constraints. I will add these numbers later.

I made a large problem to see the appearance.
```python
from qiskit_optimization.problems import QuadraticProgram

qp = QuadraticProgram()
qp.integer_var(-1, 2, 'x')
qp.integer_var(-1, 2, 'y')
qp.binary_var('u')
qp.binary_var_list(10)
qp.integer_var_list(3)
qp.continuous_var_list(3)
qp.minimize(linear={'x': 2, 'y': 3}, quadratic={('u', 'x'): -1})
qp.linear_constraint({'x': 1, 'y': -2}, '>=', 2, name='lin_GE')
qp.linear_constraint({'x': 2, 'y': -1}, '==', 1, name='lin_EQ')
qp.quadratic_constraint({'x': 1, 'u': 1}, {(3, 4): 1, (5, 6): -1}, '<=', 1, name='quad_LE')
qp.continuous_var(-1, name='z')
qp.quadratic_constraint({'x': 2, 'y': -1}, {('z', 'z'): -1}, '<=', 1)
print(qp)
```

c3a2e2d + binary w/o range
```
Problem name: 

Minimize
  - x * u + 2 x + 3 y

Subject to
  lin_GE: x - 2 y >= 2
  lin_EQ: 2 x - y == 1
  quad_LE: x3 * x4 - x5 * x6 + x + u <= 1
  q1: - z^2 + 2 x - y <= 1

Variables
  -1 <= x <= 2: integer
  -1 <= y <= 2: integer
  u: binary
  x3: binary
  x4: binary
  x5: binary
  x6: binary
  x7: binary
  x8: binary
  x9: binary
  x10: binary
  x11: binary
  x12: binary
  0 <= x13: integer
  0 <= x14: integer
  0 <= x15: integer
  0 <= x16: continuous
  0 <= x17: continuous
  0 <= x18: continuous
  -1 <= z: continuous
```

I tried another format to move names behind and consolidate binary variables.
```
# format
## constraints
  lhs sense rhs (name)

## variables
### integer or continuous
  lb <= name <= ub (type) 
 
### binary
  var1 var2 ... var_k (binary)
```

```
Problem name: 

Minimize
  -x * u + 2x + 3y

Subject to
  x - 2y >= 2  (lin_GE)
  2x - y == 1  (lin_EQ)
  x3 * x4 - x5 * x6 + x + u <= 1  (quad_LE)
  -z^2 + 2x - y <= 1  (q1)

Variables
  -1 <= x <= 2  (integer)
  -1 <= y <= 2  (integer)
  0 <= x13  (integer)
  0 <= x14  (integer)
  0 <= x15  (integer)
  0 <= x16  (continuous)
  0 <= x17  (continuous)
  0 <= x18  (continuous)
  -1 <= z  (continuous)
  u x3 x4 x5 x6 x7 x8 x9 x10 x11 x12  (binary)
```

woodsp-ibm:
How about something like this. Save repeating the type, and can easily include the count (I assume the whole sub-section would be dropped if there nothing of that sort so you would never see a count of 0)
```
Subject to:
  Linear constraints (2)
    x - 2y >= 2  'lin_GE'
    2x - y == 1  'lin_EQ'
  Quadratic constraints (2)
    x3 * x4 - x5 * x6 + x + u <= 1  'quad_LE'
    -z^2 + 2x - y <= 1  'q1'

Variables:
  Integer (5)
    -1 <= x <= 2
    -1 <= y <= 2
    0 <= x13
    0 <= x14
    0 <= x15
  Continuous (4)
    0 <= x16
    0 <= x17
    0 <= x18
    -1 <= z
  Binary (11)
    u x3 x4 x5 x6 x7 x8 x9 x10 x11 x12
```

t-imamichi:
Thanks. It looks neat. It might be simpler to put variables in the same level as constraints as follows.
```
Subject to:
  Linear constraints (2)
    x - 2y >= 2  'lin_GE'
    2x - y == 1  'lin_EQ'

  Quadratic constraints (2)
    x3 * x4 - x5 * x6 + x + u <= 1  'quad_LE'
    -z^2 + 2x - y <= 1  'q1'

  Integer variables (5)
    -1 <= x <= 2
    -1 <= y <= 2
    0 <= x13
    0 <= x14
    0 <= x15

  Continuous variables (4)
    0 <= x16
    0 <= x17
    0 <= x18
    -1 <= z

  Binary variables (11)
    u x3 x4 x5 x6 x7 x8 x9 x10 x11 x12
```

t-imamichi:
I updated the format and added four examples in https://github.com/Qiskit/qiskit-optimization/pull/184#issue-680090051

adekusar-drl (commented):
Thanks a lot for taking care of this issue. May be I'm already late, but I'll share my experience and expectation from this PR/issue cause initially I was talking about string representation of the optimization objects.

The problem of string representation arose from debugging ADMM optimizer where we create a lot of various sub-problems. Since I'm a heavy user of the PyCharm debugger, I like when objects/variables look nice in the debugger view. Apparently, they must have `__str__` and `__repr__` implemented to be represented nicely in the debugger and here we get to the present issue and potential problems in the implementation.

Pycharm debugger relies on objects' internal methods and just calls `__str__` to get a textual representation. Thus, originally an instance of `QuadraticProgram` in the debugger had a DOcplex formatted text which was not too nice, there was a long comment at the beginning, and basically you saw nothing in the debugger, only the header. OK, with the PR it is definitely better. But this is how it looks like in the debugger:
```
Problem name: ex4\n\nMinimize\n  v + w + t\n\nSubject to\n  Linear constraints (2)\n    2v + 10w + t <= 3 'cons1'\n    v + w + t >= 2  'cons2'\n\n  Binary variables (3)\n    v w t\n
```
I guess you noticed "\n" and additional spaces. Yes, I understand why it is so and there's no a silver bullet to solve this problem: either you have a nice printed view, or you have a nice view in the debugger. I'd prefer to see something like `Min: x + y + z ^ 2, s.t. N constraints`. Where `N` is just a number of constraints. Or just the objective and that's it. No additional spaces, new lines, etc. Usually it is enough in the debugger. The drawback of this approach is obvious, you see just the objective and no information on constraints, variables.

The things become more complicated when we deal with lists in the debugger, e.g. a list of constraints in a problem. When `str(a_list)` is called you get `[repr(object1), repr(object2)]`. So, in the debugger both methods, `str` and `repr` should return something meaningful and short. On other hand, an output of `repr` should be a valid python code to recreate an object. And this makes an implementation of `__repr__` complicated. There's no need to see hex numbers in the debugger, like in a case of the default implementation, but a simple text output from `repr` violates python guidelines. In case of constrains of a problem, I'd like to see just a list of text expressions of the constraints.

Well, what is a takeaway from this long text? I think, a full text representation of `QP` and probably other optimization objects may be moved to a method like `pprint` or so, while `str` should return a short string with the objective expression. In case of constraints it is not so simple. Likely `str` should return just an expression, for `repr` we have choices.

t-imamichi:
Thank you for your detailed feedback. It is really useful. You appear in time :-)
I will check pycharm debugger and get back to you because I never tested this branch with pycharm debugger.

As for `repr`, it is preferable to be a valid python code, but we are allowed to use `<QP ....>` like expression according to the following statement.

> If this (a valid Python expression) is not possible, a string of the form <...some useful description...> should be returned.

https://docs.python.org/3/reference/datamodel.html#object.__repr__

adekusar-drl:
That's right, I think `<... some useful description ...>` is the best choice for `__repr__`, sort of compromise.

t-imamichi:
I tried Pycharm's debug mode. It uses `repr` for a list of objects, but uses `str` for a single object.
image

I think printing multiple lines makes the debug screen complicated. I think `str` should be one line as well as `repr`, but `str` can be more informative because `str` is not used for a list of objects.

How about the following idea?

Idea 1:
- str: `<QuadraticProgram at 0x7fb8e20c0520: name='problem 2', obj_sense=maximize, num_variables=2, num_constraints=1>`
- repr: `<QuadraticProgram at 0x7fb8e20c0520: name='problem 2'>`

Idea 2:
- str: `<QuadraticProgram:  name='problem 2', obj_sense=maximize, num_variables=2, num_constraints=1>`
- repr: `<QuadraticProgram: name='problem 2'>`

I'm wondering about the id part `0x7fb8e20c0520`. It is useful to check the identity of objects. But, it may not be so informative in general.

adekusar-drl:
> 
> 
> I tried Pycharm's debug mode. It uses `repr` for a list of objects, but uses `str` for a single object.

PyCharm uses `str` for all objects, but `list.__str__()` uses `repr()` on the list's elements. That's why it is slightly messy.
 
> I think printing multiple lines makes the debug screen complicated. I think `str` should be one line as well as `repr`, but `str` can be more informative because `str` is not used for a list of objects.

Exactly, this is what I realized some time ago with the original implementation!

> 
> How about the following idea?
> 
> Idea 1:
> 
>     * str: `<QuadraticProgram at 0x7fb8e20c0520: name='problem 2', obj_sense=maximize, num_variables=2, num_constraints=1>`
> 
>     * repr: `<QuadraticProgram at 0x7fb8e20c0520: name='problem 2'>`

I'd like to see objective and sense in the `str`. I guess problem names are not that important, but this is really subjective.

> 
> Idea 2:
> 
>     * str: `<QuadraticProgram:  name='problem 2', obj_sense=maximize, num_variables=2, num_constraints=1>`
> 
>     * repr: `<QuadraticProgram: name='problem 2'>`
> 
> 
> I'm wondering about the id part `0x7fb8e20c0520`. It is useful to check the identity of objects. But, it may not be so informative in general.

yeah, hex part may be convenient in some cases, but pretty rarely, I believe.

> 
> I implemented `to_str` function for detailed print out of a problem. We can discard it or leave it as a translator by renaming it `prettyprint`.

woodsp-ibm:
I see there is a quite old feature request here for it to use repr https://youtrack.jetbrains.com/issue/PY-35529 as an option. States VSCode does.

While I can see this is an aspect of designing things most users are not going to be debugging like this - hopefully! I can see the argument for having a more compact str than the multiline one originally proposed that as suggested it most likely better off as a separate method that has to be more explicitly called - such as `prettyprint` or `detailed_info` that returns a formatted string that you can print, save to disk or whatever.

I do think str should be more human friendly, which is what its supposed to be - even if its what PyCharm debugger shows. Something more like this perhaps (since parts ought to be obvious I dont think it needs to be name= etc but more informal like below.

> QuadraticProgram: 'my_problem', x+2yx > 0, maximize, 2 vars, 3 constraints` 

Maybe the objective expression can become very long hence perhaps that's not there. I do not think hex id is needed in str - its in the default repr so there I think is fine so you can see if its the same object instance or not.

t-imamichi:
@adekusar-drl @woodsp-ibm I tried to implement `repr` and `str` of variables, expressions, constraints and objectives as well as QP. I think this new text representation makes Pycharm's debugger more informative. I would like to hear your feedback.

Basic ideas:
- renamed `to_str(QP)->str` with `prettyprint(QP)->str` as @woodsp-ibm suggested https://github.com/Qiskit/qiskit-optimization/pull/184#issuecomment-890095478
- `repr`: `<class_name: prettyprint of expression>`
  - I followed the Python's convention `<...some useful description...>`. I added a class name because it is used for a list in Pycharm debugger and Pycharm debugger shows just `list` for a list. So, class names might be useful.
  - I added truncation with length 50 similar to [quantum_info's Pauli](https://qiskit.org/documentation/stubs/qiskit.quantum_info.Pauli.html). If the expression is too long, it is truncated, e.g., "x + y +...". It might be useful because Pycharm uses `repr` for a list.
- `str`: `prettyprint of expression` (no class name, no truncation)
  - I omit the class name because Pycharm's variable pane shows the class name
  - I disable truncation for `str` because Pycharm uses `str` for a single object. I think a long string does not interrupt other objects in the pane.
 
Details:
- Variable
  - `repr`: <Variable: {name} binary> or <Variable: {lowerbound} <= {name} <=  {upperbound} {variable type}>
    - I didn't truncate it because truncation might remove upper bound 
  - `str`: "{name} (binary)"  or "{lowerbound} <= {name} <=  {upperbound} ({variable type})"
- LinearExpression or QuadraticExpression or QuadraticObjective
  - `repr`: <{class name}: {prettyprint of expression with truncation}>
  - `str`: "{prettyprint of expression}"
- LinearConstraint or QuadraticConstraint
  - `repr`: <{class name}: {prettyprint of expression with truncation} '{constraint name}'>
  - `str`: "{prettyprint of expression} '{constraint name}'"
- QuadraticProgram
  - `repr`: <QuadraticProgram: {sense} {prettyprint of objective function with truncation}, {x} variables, {y} constraints, '{name}'>
  - `str`:  "{prettyprint of objective function with truncation} ({x} variables, {y} constraints, '{name})'

An example:
```python
from qiskit_optimization.problems import QuadraticProgram
from qiskit_optimization.translators.prettyprint import prettyprint

qp = QuadraticProgram('problem 1')
qp.integer_var(-1, 2, 'x')
qp.integer_var(-1, 2, 'y')
qp.continuous_var(-1, name='z')
qp.binary_var('u')
qp.binary_var('very_long_variable_' * 3)
qp.binary_var_list(10)
qp.integer_var_list(3)
qp.continuous_var_list(3)
qp.minimize(constant=3, linear={'x': 2, 'y': 3}, quadratic={('u', 'x'): -1})
qp.linear_constraint({'x': 1, 'y': -2}, '>=', 2, name='lin_GE')
qp.linear_constraint({'x': 2, 'y': -1}, '==', 1, name='lin_EQ')
qp.quadratic_constraint({'x': 1, 'u': 1}, {(3, 4): 1, (5, 6): -1}, '<=', 1, name='quad_LE')
qp.quadratic_constraint({'x': 2, 'y': -1}, {('z', 'z'): -1}, '<=', 1)
print(prettyprint(qp))
```

output:
```
Problem name: problem 1

Minimize
  -x * u + 2x + 3y + 3

Subject to
  Linear constraints (2)
    x - 2y >= 2  'lin_GE'
    2x - y == 1  'lin_EQ'

  Quadratic constraints (2)
    u * very_long_variable_very_long_variable_very_long_variable_ - x5 * x6 + x + u <= 1  'quad_LE'
    -z^2 + 2x - y <= 1  'q1'

  Integer variables (5)
    -1 <= x <= 2
    -1 <= y <= 2
    0 <= x15
    0 <= x16
    0 <= x17

  Continuous variables (4)
    -1 <= z
    0 <= x18
    0 <= x19
    0 <= x20

  Binary variables (12)
    u very_long_variable_very_long_variable_very_long_variable_ x5 x6 x7 x8 x9 x10 x11 x12 x13 x14
```

Pycharm's debug window
image

qiskit_optimization/problems/variable.py
```diff
@@ -148,3 +148,17 @@ def as_tuple(self) -> Tuple[str, Union[float, int], Union[float, int], VarType]:
             variable type.
         """
         return self.name, self.lowerbound, self.upperbound, self.vartype
+
+    def __repr__(self):
+        if self._vartype == VarType.BINARY:
+            return f"<Variable: {self._name} {self._vartype.name.lower()}>"
+        lowerbound = "" if self._lowerbound == -INFINITY else f"{self._lowerbound} <= "
+        upperbound = "" if self._upperbound == INFINITY else f" <= {self._upperbound}"
+        return f"<Variable: {lowerbound}{self._name}{upperbound} {self._vartype.name.lower()}>"
```
woodsp-ibm, line 157: Since repr is just str plus the Variable text you not do repr much easier as something like
```
return f"<Variable: {str(self)}>"
```
to save replicating the logic.

woodsp-ibm:
In general it looks fine to me. I see its failing due presumably to the new code added which does not yet have unit test coverage so the overall percentage dropped.

t-imamichi:
Yes, I still don't add any unit test yet. If we agree on the format, I will write unit tests.

adekusar-drl:
@t-imamichi Looks very nice to me! Thanks a lot for your efforts!

t-imamichi:
Thank you for your comments. I will write unit tests with this format.

qiskit_optimization/problems/variable.py
t-imamichi: Thanks. I will do it.

t-imamichi:
I'm working on adding unit tests gradually in another branch not to invoke CI.
https://github.com/t-imamichi/qiskit-optimization/compare/repr...t-imamichi:repr-test

t-imamichi:
TODO
- [] update PR description
- [ ] add reno
- [ ] add `QuadraticProgram.prettyprint`
- [ ] update tutorials

garrison:
One thing I've noticed about switching from `__repr__` to `__str__` is that Jupyter notebooks render the `__repr__` by default, rather than the `__str__`.  So instead of placing the expression at the end of the cell (e.g. `results`), you must explicitly display it (e.g. `print(results)`) if you want the human-readable version.

I think this might be part of what you mean by the following point but wanted to mention it explicitly:

> Check tutorials that use the current `__repr__`

t-imamichi:
Thank you for the information. I should check how jupyter renders objects as well as PyCharm and vscode do.

t-imamichi:
This PR is almost ready, but it gets larger than I expected.
So I will split this into three PRs to make review easier.

- [ ] adds `prettyprint`
- [ ] adds `__str__` and `__repr__` to problem related objects except `QuantumProgram`
- [ ] adds `__str__` and `__repr__` to `QuantumProgram` and update all tutorials

PR #185, opened 2021-06-30
Fix a warning by pylint 2.9.0

t-imamichi:
### Summary

Pylint 2.9.0 is released today. This PR fixes the following warning raised by pylint 2.9.0.
```
************* Module qiskit_optimization.algorithms.warm_start_qaoa_optimizer
qiskit_optimization/algorithms/warm_start_qaoa_optimizer.py:95:8: C0206: Consider iterating with .items() (consider-using-dict-items)
```


mtreinish (approved):
LGTM, we should really be pinning pylint and astroid to a fixed version in the requirements list to avoid random failures like this on new releases. But we can do that in a follow up after fixing the failure.

t-imamichi:
Thank you for reviewing it. We discussed pylint version in our meeting today and decided to use the latest version because the latest version raises only a few warnings.

PR #186, opened 2021-06-30
Upgrade pylint to 2.8.3

ikkoham:
### Summary

Terra's pylint is upgraded to 2.8.3. 
https://github.com/Qiskit/qiskit-terra/pull/6317

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,7 @@
 coverage>=4.4.0
 matplotlib>=2.1
 black==21.4b2
-pylint>=2.7.1
+pylint>=2.8.3
 pylatexenc>=1.4
 stestr>=2.0.0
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 coverage>=4.4.0
 matplotlib>=2.1
 black==21.4b2
-pylint>=2.7.1
+pylint>=2.8.3
```
mtreinish, line 4: We probably should pin this (and astroid) to an exact version to avoid spurious failures caused by new releases

ikkoham: Thanks. The version is pinned by `constraint.txt`. I forgot to upgrade...

manoelmarques (approved):

PR #189, opened 2021-07-01
Fix a couple of notebook titles and list formatting

woodsp-ibm:
A couple of older notebook titles had the original italic style, so I changed that so they are consistent with how the rest are now, so that they come out in the documentation index looking consistent.

Also fixed list formatting. With jupyter notebook a list works without having a blank line preceding it. But now we are converting these and creating html using Sphinx we need a blank line otherwise all the text just wraps into a single paragraph.


docs/tutorials/06_examples_max_cut_and_tsp.ipynb
```diff
@@ -136,13 +136,396 @@
...
+      "image/svg+xml": [
```
t-imamichi, line 141: I don't think you need to modify the image. Btw, I turned off the svg format because it causes a lot of line changes. I'm wondering why your commit still includes svg part. On my local mac, the svg part is not generated.

woodsp-ibm: I have no idea - I would have to look at jupyter notebook and see if there are any settings in this regard that it creates this when it saves, All I did was to edit the title and save it, nothing more.

t-imamichi: This might be due to notebook version. I committed this tutorial with notebook 6.3 or 6.4. What is your version?

woodsp-ibm: Seems quite a bit older 4.4

t-imamichi: If you update the notebook version, the diff may get smaller. But, I approve this change because the it's not so important.

t-imamichi (approved):

docs/tutorials/06_examples_max_cut_and_tsp.ipynb
woodsp-ibm: Yes, it might be a good idea for me to update - as long as I don't end up with all those random ids in the file. I think they may come from nbformat - what version is yours?

t-imamichi: Mines are as follows.
```
notebook                      6.4.0
nbconvert                     6.1.0
nbformat                      5.1.3
```

woodsp-ibm: FYI: I updated to those versions. Loaded it up and saved it and its unchanged from before; made changes to it just in case but its still the same for me in regards the image sections. So I tried...

t-imamichi: Thank you for trying the latest version.

PR #190, opened 2021-07-01
Getting started: fix typo

epelaaez:
### Summary
Changed "qiskit[optimizationg]" to "qiskit[optimization]" in getting started file.


woodsp-ibm (approved):
Thanks for noticing and fixing this. Normally our spell checker, that we run on the build, picks up mistakes like that. Maybe since its in a quoted section it did not do so. Either way thanks again for doing this!

t-imamichi (approved):

PR #197, opened 2021-07-08
Fix the link to the release note of Qiskit optimization

t-imamichi:
### Summary

Fix the link to the release note of Qiskit. main -> master


CONTRIBUTING.md
```diff
@@ -223,7 +223,7 @@ it has been tagged::
 At release time ``reno report`` is used to generate the release notes for the
 release and the output will be submitted as a pull request to the documentation
 repository's [release notes file](
-https://github.com/Qiskit/qiskit/blob/main/docs/release_notes.rst)
+https://github.com/Qiskit/qiskit/blob/master/docs/release_notes.rst)
```
woodsp-ibm, line 226: Since this is about optimization, and the release notes are now separate from the main ones, would it be better for this to point here https://qiskit.org/documentation/optimization/release_notes.html

t-imamichi: Thanks. I fixed it. The file is currently almost empty, but we will accumulate release notes there. Is it correct?

woodsp-ibm: It never changes. The qiskit repo one is the one where the various notes from Terra, Aer, Aqua etc are aggregated, Since Aqua was split into application repos, and had its own independent release notes, which is built out when the docs are built/published, in looking more closely I think that section has no real meaning nowadays. Its more likely a leftover from how things were - @manoelmarques any thoughts - just remove this or reword it. In the end its an aspect that most contributors will not see/know/care about I think.

t-imamichi: @manoelmarques Do you have any idea of what to do with this link?

woodsp-ibm: My take above was to remove the whole paragraph - I am not sure what value it has to most contributors. Sure they need to know how to create a release note - and perhaps someone might like to see how it finally looks later.

If you really wanted to say something maybe its more like
```
At release time ``reno report`` is used to generate the release notes for the
release and the output is used to create the release notes that form part of the documentation
published here https://qiskit.org/documentation/optimization/release_notes.html
```

t-imamichi: Thanks. I removed the paragraph.

woodsp-ibm (approved):

PR #203, opened 2021-07-12
Remove random_graph method from GraphOptimizationApplication

### LINKED ISSUE #201: GraphOptimizationApplication, single responsibility principle

the static method `random_graph` in the `GraphOptimizationApplication` 
violates the single responsibility principle. Please create some helper and move it there.

a-matsuo:
### Summary
Fix #201.

### Details and comments
Remove `random_graph` method from `GraphOptimizationApplication`.

Btw, do we need to deprecate before removing?


t-imamichi:
Yes, we should mark it deprecated first.

a-matsuo:
I added a deprecation message. I was not sure what the next version number will be, so I just put 0.3.0 as the next version. We might need to change it when the next version number is fixed.
```
DeprecationWarning: The random_graph method is deprecated as of version 0.3.0 and will be removed no sooner than 3 months after the release.
```

qiskit_optimization/applications/graph_optimization_application.py
```diff
@@ -94,9 +95,9 @@ def graph(self) -> nx.Graph:
         return self._graph
 
     @staticmethod
+    @deprecate_method("0.3.0", DeprecatedType.FUNCTION)
```
woodsp-ibm, line 98: Version 0.3.0 is correct. As soon as the release goes out we update the main branch version file so aside from a very short period at release time, it should be always be the next version https://github.com/Qiskit/qiskit-optimization/blob/207d3215cca04f8006d65644a644c85adc9611c7/qiskit_optimization/VERSION.txt#L1

Can we include that they should just use `nx.gnm_random_graph` directly as a replacement

a-matsuo: Oh, I didn't know we have such a file. Thank you!
Also, I updated to use `nx.gnm_random_graph` directly.

woodsp-ibm: > Can we include that they should just use nx.gnm_random_graph directly as a replacement

Sorry I did not mean a code change to directly return the value, I was thinking that in the deprecation we should tell the user that they can use nx.gnm_random_graph as a replacement (i.e. they can directly call that method rather than using this, now deprecated, method)

a-matsuo: I see xD I added additional message to guide users to use NetworkX directly. It's now like this.
```
DeprecationWarning: The random_graph method is deprecated as of version 0.3.0 and will be removed no sooner than 3 months after the release. Instead use the netowrkx.gnm_random_graph function in NetworkX, directly.
```

qiskit_optimization/applications/graph_optimization_application.py
```diff
@@ -95,7 +95,7 @@ def graph(self) -> nx.Graph:
         return self._graph
 
     @staticmethod
-    @deprecate_method("0.3.0", DeprecatedType.FUNCTION, "netowrkx.gnm_random_graph", "directly")
+    @deprecate_method("0.3.0", DeprecatedType.FUNCTION, "netowrkx.gnm_random_graph", "in NetworkX, directly")
```
woodsp-ibm, line 98:
```suggestion
    @deprecate_method("0.3.0", DeprecatedType.FUNCTION, "networkx.gnm_random_graph", "in NetworkX, directly")
```

woodsp-ibm: Small spelling correction for networkx.

woodsp-ibm (approved):

t-imamichi (approved):

PR #150, opened 2021-05-28
[WIP] Replace the internal networkx with retworkx

t-imamichi:
### Summary

Replace the internal networkx data structure of applications with retworkx for better scaling in the future.

Fixes #147

PR #208, opened 2021-07-22
application: Sherrington Kirkpatrick (SK) model

X-Libor:
### Summary
I added the Sherrington Kirkpatrick (SK) model to the applications part of qiskit-optimization module.


qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+from ..algorithms import OptimizationResult
+from ..problems.quadratic_program import QuadraticProgram
+from ..translators import from_docplex_mp
+
+
+class SKModel(OptimizationApplication):
+    """Optimization application for the "Sherrington Kirkpatrick (SK) model" [1].
+
+    The SK Hamiltonian over n spins is given as:
+        H(x)=-1/\sqrt{n} \sum_{i<j} w_{i,j}x_ix_j, where x_i\in\pm 1
```
dlasecki, line 29: Using the math formatting would be useful here to render the equation nicely.

qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+    and w_{i,j}\in\pm 1 are chosen independently and uniformly at random.
```
dlasecki, line 30: Using the math formatting would be useful here to render the equation nicely.

qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+
+        objective = mdl.sum(
+            -1 / np.sqrt(self._n) * self._graph.edges[i, j]["weight"] * (2 * x[i] - 1) * (2 * x[j] - 1)
+            for i, j in self._graph.edges
+        )
+        # we converted the standard H(x)=-1/\sqrt{n} \sum w_{ij}x_ix_j, where x_i\in\pm 1 to binary.
+
+        mdl.minimize(objective)
+        op = from_docplex_mp(mdl)
+        return op
```
dlasecki, line 79:
```suggestion
        return from_docplex_mp(mdl)
```

qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def interpret(self, result: Union[OptimizationResult, np.ndarray]) -> List[List[int]]:
+        """Interpret a result as energy and configuration of spins
```
dlasecki, line 82:
```suggestion
        """Interpret a result as energy and configuration of spins.
```

qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+        self.new_instance()
+
+    def new_instance(self) -> None:
+        """Generate a new instance of the SK model"""
+        for _, _, d in self._graph.edges(data=True):
+            d['weight'] = self._rng.choice([-1, 1])
+
+    def to_quadratic_program(self) -> QuadraticProgram:
+        """Convert an SK model problem instance into a
+        :class:`~qiskit_optimization.problems.QuadraticProgram`
```
dlasecki, line 60:
```suggestion
        :class:`~qiskit_optimization.problems.QuadraticProgram`.
```

qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+            rng: numpy pseudo-random number generator
+        """
+        self._rng = rng
+        self._n = n
+        self._graph = nx.complete_graph(self._n)
+
+        self.new_instance()
+
+    def new_instance(self) -> None:
+        """Generate a new instance of the SK model"""
```
dlasecki, line 54:
```suggestion
        """Generate a new instance of the SK model."""
```

qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+
+        mdl.minimize(objective)
+        op = from_docplex_mp(mdl)
+        return op
+
+    def interpret(self, result: Union[OptimizationResult, np.ndarray]) -> List[List[int]]:
+        """Interpret a result as energy and configuration of spins
+
+        Args:
+            result : The calculated result of the problem
```
dlasecki, line 85:
```suggestion
            result : The calculated result of the problem.
```

qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+
+        Returns:
+            [energy, configuration of spins]
+        """
+        configuration = list(map(lambda x: 2 * x - 1, self._result_to_x(result)))
+        return [result.fval, configuration]
+
+    @property
+    def graph(self) -> nx.Graph:
+        """Getter of the graph
```
dlasecki, line 95:
```suggestion
        """Getter of the graph.
```

qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+        Returns:
+            A graph for a problem
```
dlasecki, line 97:
```suggestion
            A graph for a problem.
```
Check all other cases.

qiskit_optimization/applications/SK_model.py
```diff
@@ -0,0 +1,104 @@
...
+    def to_quadratic_program(self) -> QuadraticProgram:
+        """Convert an SK model problem instance into a
+        :class:`~qiskit_optimization.problems.QuadraticProgram`
+
+        Returns:
+            The :class:`~qiskit_optimization.problems.QuadraticProgram` created
+            from the SK problem instance.
+        """
+        mdl = Model(name="SK-model")
+        x = {
```
dlasecki, line 67: Can we think of a more descriptive name?

dlasecki (changes requested):

qiskit_optimization/applications/SK_model.py
X-Libor: I took it from MaxCut. There is `name="Max-cut"`.

dlasecki: Sorry, I had `x` in mind.

.gitignore
```diff
@@ -111,6 +111,7 @@ ENV/
 env.bak/
 venv.bak/
 
+
```
woodsp-ibm, line 114: Seems you could rollback the change on .gitignore - an extra blank line?

test/applications/test_SK_model.py
```diff
@@ -0,0 +1,81 @@
...
+            variables=op.variables,
+            status=OptimizationResultStatus.SUCCESS,
+        )
+
+    def test_to_quadratic_program(self):
+        """Test to_quadratic_program"""
+        problem = SKModel(self._n, np.random.RandomState(self._seed))
+        qp = problem.to_quadratic_program()
+
+        # Test name
```
woodsp-ibm, line 51: When there are in essence multiple independent tests its good to use the with self.subTest('test description') construct so that the entire test continues even if that fails. As it stands here any assertion will end the overall test. I imagine for instance that constraint could still be correct even if the objective messed up. I'll assume that under where you say test objective that these are most likely not independent enough to warrant that. BTW isn't places=7 the default unless we need something else I think its better not to have it - especially as a positional argument instead of using a named argument where.

woodsp-ibm:
In regards of the code - if you run `make black` in youre repo clone root then it will format the code according to how its checked (CI checks the format via black but does not change things) We use black nowadays to ensure consistent style rather so the tool can just be run - rather than using style checks and expecting the code to be manually fixed.

qiskit_optimization/applications/SK_model.py
X-Libor: It is a vector of binary variables that is called `x`. Do you think it requires a different name?

test/applications/test_SK_model.py
woodsp-ibm: Check this test case out as an example of subTest - it has the with for context setting and the code then comprising that subtest is indented https://github.com/Qiskit/qiskit-optimization/blob/e3b878ee697cc7ebba8867d3a20394a404642c0d/test/converters/test_linear_inequality_to_penalty.py#L39-L41

.gitignore
X-Libor: Can you please advice me how to do so?

test/applications/test_SK_model.py
X-Libor: Thank you. Done. Please check.

qiskit_optimization/applications/SK_model.py
X-Libor: Can you please advice me what do you expect? Thank you.

.gitignore
woodsp-ibm: The simplest way in this case would just be to delete the blank line I think. That way when you commit it again I think the total number of files changed will not include this one since there is no net change.

test/applications/test_SK_model.py
woodsp-ibm: That looks better now - thanks.

qiskit_optimization/applications/SK_model.py
woodsp-ibm: I think he is suggesting to use the math formatting like this example here does https://github.com/Qiskit/qiskit-optimization/blob/4347110991ba0ff656d20cbf50aaf1618ebd19b8/qiskit_optimization/problems/quadratic_program.py#L1043-L1049 which you can see how Sphinx formats it and its finally rendered here https://qiskit.org/documentation/optimization/stubs/qiskit_optimization.QuadraticProgram.from_ising.html

woodsp-ibm:
For spelling issues, names are often the case where they are not in the standard dictionary, we have a custom dictionary file here https://github.com/Qiskit/qiskit-optimization/blob/main/.pylintdict where you just add the word(s) - lowercase and in order alphabetically so its easier to manage - so if the word is correctly spelled please just add it to that dictionary file and commit the change with this PR.

qiskit_optimization/applications/SK_model.py
X-Libor: Thank you. Done. Please check.

qiskit_optimization/applications/SK_model.py
X-Libor: Thank you. Done. Please check.

.gitignore
X-Libor: Thank you. fixed.

woodsp-ibm:
To have the documentation come out and have the class exposed it should be added to the applications init file here https://github.com/Qiskit/qiskit-optimization/blob/main/qiskit_optimization/applications/__init__.py

Since you fixed the spelling I assume you have figured you look at the build Details of a given job to see what is failing - often its the same across 3.6, 3.7 etc where we test under the different Python versions.

qiskit_optimization/applications/SK_model.py
dlasecki: Yes, this is what I had in mind. Looks good.

qiskit_optimization/applications/SK_model.py
dlasecki: Maybe `decision_vars` or sth along these lines?

X-Libor: I have couple reasons why `x` is better:
1) it makes the objective function more readable.
2) it seems to be a convention, please see other applications max_cut, knapsack, graph_partition, ... if you want to change it it should be refactored in the whole module.
3) I like `x`. :)) ok, it is better aligned to what you see in the literature then `decision_vars`. Also it is an internal variable in this function.

qiskit_optimization/applications/SK_model.py
```diff
@@ -36,7 +36,7 @@ class SKModel(OptimizationApplication):
         https://arxiv.org/abs/1211.1094
     """
 
-    def __init__(self, n: int, rng: np.random.RandomState = np.random):
+    def __init__(self, n: int, rng: np.random.RandomState = np.random.RandomState(0)):
```
woodsp-ibm, line 39: Since this application is all around random behavior I guess its ideal that the random behavior here is independent. What we have otherwise is a random generator in algorithm_globals (in qiskit-terra utils folder now) that we use in general. This provides random behavior but we can see its behavior to have reproducible results. If this is independently random - but I see it truly isn't by default as you seed it. Might it be better to go over to the common generator we have - or if you want this independent allow an optional Generator or a seed to passed in. By default you can used np.random.default_rng() and if a seed is passed pass that to the default_rng() otherwise the user supply a Generator of their choosing. I think RandomState is more associated to the legacy random aspects of numpy.

X-Libor: Thank you. Ok, I don't understand rng's in python.
I rea that there might be issues with reproducibility and RandomState is a good way how to deal with it https://stackoverflow.com/a/37224605
What do you suggest?

dlasecki (approved):

t-imamichi:
Could you add `SKModel` to the docstring of `__init__.py` of applications?

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,110 @@
...
+        :class:`~qiskit_optimization.problems.QuadraticProgram`.
+
+        Returns:
+            The :class:`~qiskit_optimization.problems.QuadraticProgram` created
+            from the SK problem instance.
+        """
+        mdl = Model(name="SK-model")
+        x = {
+            i: mdl.binary_var(name="x_{0}".format(i)) for i in range(self._graph.number_of_nodes())
+        }
```
t-imamichi, line 72:
```suggestion
        x = mdl.binary_var_list(self._graph.number_of_nodes())
```

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,110 @@
...
+    and :math:`w_{i,j}\in\{\pm 1\}` are chosen independently and uniformly at random.
+    Notice that there are other variants e.g., with :math:`w_{i,j}` chosen from the normal distribution
+    with mean 0 and variance 1.
+
+    References:
+        [1]: Dmitry Panchenko. "The Sherrington-Kirkpatrick model: an overview",
+        https://arxiv.org/abs/1211.1094
+    """
+
+    def __init__(self, num_of_sites: int, rng: np.random.RandomState = np.random.RandomState(0)):
```
t-imamichi, line 42: Could you use `Random Generator` instead of legacy `RandomState`? https://numpy.org/doc/stable/reference/random/generator.html

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,110 @@
...
+        """Getter of the graph.
+        Returns:
+            A graph for a problem.
+        """
+        return self._graph
+
+    def _set_rng(self, rng):
+        self._rng = rng
+
+    rng = property(fset=_set_rng)
```
t-imamichi, line 110: If you want to implement a setter, could you use the following idiom with a docstring as in the python doc?
https://docs.python.org/3/library/functions.html?highlight=setter#property
```
    @x.setter
    def x(self, value):
        """ Documentation comes here
        """
        self._x = value
```

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,110 @@
...
+    def interpret(self, result: Union[OptimizationResult, np.ndarray]) -> List[int]:
+        """Interpret a result as configuration of spins.
+
+        Args:
+            result : The calculated result of the problem.
+
+        Returns:
+            configuration of spins
+        """
+        configuration = list(map(lambda x: 2 * x - 1, self._result_to_x(result)))
```
t-imamichi, line 96:
```suggestion
        configuration = [2 * x - 1 for x in self._result_to_x(result)]
```

qiskit_optimization/applications/sk_model.py
woodsp-ibm: I guess this is a similar conversation to above here https://github.com/Qiskit/qiskit-optimization/pull/208#discussion_r675855471

My question first was did this need to be an independent rng. It could internally just use utils\algorithm_globals rng from Terra, as is done in other places in application so this can be seeded from the one central place for over all reproducibility.

if it does need to be independent the next question is do we really need to pass different generators in. Yes they have different properties around the generation but is that needed/wanted? Maybe setting some Optional int seed and having it use default_rng, like algorithm_globals may suffice. If we want generator then maybe a Optional Union of Generator & int whereby a generator can be provided or more simply a seed for default_rng (i.e. do a blend). I cannot help but feel that it getting more complex though and does it really warrant going that far with rng options?

qiskit_optimization/applications/sk_model.py
woodsp-ibm: I am not sure why we would want an rng setter - isn't setting/influencing it via constructor enough? What does this even do since the constructor builds using the rng and I do not see it used afterwards that a setter would have any value.

qiskit_optimization/applications/__init__.py
```diff
@@ -74,4 +75,5 @@
     "Tsp",
     "VehicleRouting",
     "VertexCover",
+    "SKModel"
```
woodsp-ibm: It also needs adding above to the other list that will form the docs. If you keep it in alphabetic order there, which will help the reader find it easier, then it will become an entry between current lines 43 & 44.

qiskit_optimization/applications/sk_model.py
X-Libor: Hi, yes, we need to have it reproducible. So in any case it should return deterministic pseudo-random numbers.

qiskit_optimization/applications/sk_model.py
X-Libor: I thought it might be useful to be able to change the pseudorandom number generator.
It will be used in the next call of `new_instance()` which generates a new disorder of the SK model.
Obviously you can create a new instance of the `SKModel` class and start with a new seed.
And maybe it is not that useful after all.

qiskit_optimization/applications/sk_model.py
X-Libor: Hi, yes, we need the rng to be completely independent and produce deterministic pseudo-random numbers.
Yes it is a similar comment as [https://github.com/Qiskit/qiskit-optimization/pull/208#discussion_r675855471](https://github.com/Qiskit/qiskit-optimization/pull/208#discussion_r675855471).

What do you suggest? Objective is we want deterministic pseudo-random numbers independent on anything else.

qiskit_optimization/applications/sk_model.py
X-Libor: I guess that this should be then refactored also for other problems in this module.

qiskit_optimization/applications/__init__.py
X-Libor: Thank you. Done.

qiskit_optimization/applications/sk_model.py
t-imamichi: I think it's confusing to provide rng setter. If users want to change rng, they can create a new object with a new rng generator.

qiskit_optimization/applications/sk_model.py
t-imamichi: rng is essentially used to set values of edges and the graph is always complete graph. So I would recommend to receive edge weight as `np.ndarray` with shape = (num_of_sites, num_of_sites). Then, users can generate any edge weight as they want. The current design allows only random edge weights.

Related to the recommendation above, I also recommend to use np.ndarray as the internal data structure because the graph is complete.

qiskit_optimization/applications/sk_model.py
X-Libor: ok. I take it out.

qiskit_optimization/applications/sk_model.py
X-Libor: I think that providing edge weights as an array should be a side option only, not the only one.
Ultimately the model/problem is about the random distortions.

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,108 @@
...
+            num_of_sites: number of sites
+            rng: numpy pseudo-random number generator
+        """
+        self._rng = rng
+        self._num_of_sites = num_of_sites
+        self._graph = nx.complete_graph(self._num_of_sites)
+
+        self.new_instance()
+
+    def new_instance(self) -> None:
```
t-imamichi, line 56: I have a concern about this public method. Users can generate different SK model by giving different values to the constructor. Why do you think this method is necessary no top of the constructor? @X-Libor

qiskit_optimization/applications/sk_model.py
X-Libor: Regarding not having it as a graph: It is a possibility.
However, having it as a graph have some advantages:
* For some test examples it might be good to be able to draw it with displayed edge weights.
* It is related to MaxCut on random graphs. So by design it is a graph problem. (BTW some variant of SK can have also edge weights 0)

It is subtle, I am not sure where to draw the line. So I picked this as an easy solution.

BTW: recirq has it also represented as a graph.

I would personally stick with the graph representation.

qiskit_optimization/applications/sk_model.py
X-Libor: In general you are interested in typical behavior for this model. Hence, you are going to generate large number of instances of the same number of sites with different --- random --- distortions.

Maybe the whole class is more a factory for SK model instances.

qiskit_optimization/applications/sk_model.py
t-imamichi: If this is a graph problem, could you inherit `GraphOptimizationProblem` instead of `OptimizationProblem`?
Note that `random_graph` is deprecated by #203. But, you can make a utility function as follows, e.g.,
```python
class SKModel(GraphOptimizationProblem):
    def __init__(self, ...):
        ...

    @staticmethod
    def random_instance(num_nodes, rng: Generator) -> SKModel:
        edge_weights = rng.uniform(...)
        return SKModel(edge_weights)
```

This is also a reply to this comment. https://github.com/Qiskit/qiskit-optimization/pull/208#discussion_r677332636

qiskit_optimization/applications/sk_model.py
t-imamichi: You don't need to overwrite an existing object to try another instance. I suggest to generate a new object with a new random number generator as suggested https://github.com/Qiskit/qiskit-optimization/pull/208#discussion_r677635115

qiskit_optimization/applications/sk_model.py
X-Libor: It is not a general graph problem. That is the reason I don't inherit from that class. I don't want users to be able to set the graph and so..

Yes that was my request to remove that function. I considered that before.

I think it should contain a graph but not inherit from the GraphOptimizationProblem if it is what it is right now.

qiskit_optimization/applications/sk_model.py
X-Libor: then you would need to keep the rng somewhere else to have it reproducible.
 :(

I also dont want that some other part of code can change the state of the rng and make it not reproducible.

woodsp-ibm: I guess I am still not clear. When you say 100000 instances these are all different I assume - but reproducible as a set given the same random number generator. Hence you want the same generator to be used by different instances otherwise if they have their own generator and its seeded they will all be the same assuming the same seed. Of course you can do with 100000 different fixed seed values - but maybe thats too much. On the other hand how many problem instances do you normally expect users to play with?

X-Libor: Yes, in that example I meant 100000 random instances. Reproducible: they would be generated by one independent seeded random number generator. Now they are produced by invoking the new_instance method, which changes the weights.

Maybe this should be implemented through some design pattern like factory.

qiskit_optimization/applications/sk_model.py
t-imamichi: Could you explain how it is not a general graph problem? Then, we can come up with ideas.

qiskit_optimization/applications/sk_model.py
t-imamichi: If you are allowed to destroy an existing object by `new_instance` as `sk_model.new_instance()`, it is essentially same as replacing an object with a new object with a different edge weights as `sk_model = SKModel(new_edge_weights)`. Generating an object from edge weights is intuitive from reproducibility point of view.

qiskit_optimization/applications/sk_model.py
X-Libor: I dont want users to input graph or weights. The model should stay as it is.

qiskit_optimization/applications/sk_model.py
X-Libor: I dont want users to pick edge weights. They would possibly not do it in the way model is intended.

t-imamichi: What if users want to solve a SK model generated by their own rng? I believe we need to make edge weights as an input of the constructor and provide a utility static method to generate an instance with numpy's rng similar to `new_instance`.

t-imamichi: If maxcut is sufficient to represent the problem, we don't need to provide a new class. It seems enough to make a `generate_sk_model` to `Maxcut` class.

X-Libor: > What if users want to solve a SK model generated by their own rng?
Ok. Nevertheless if they want this to be repeatable then it has to be an independent rng.

You mention in the docstring that
> 
> > Notice that there are other variants e.g., with :math:`w_{i,j}` chosen from the normal distribution
This is not related to the previous.

> So, some users may want to set edge weights generated from a normal distribution.
Yes, it can be generalized in some other iteration. Right now this is about models with weights \pm 1 with uniform independent.
 
> I believe we need to make edge weights as an input of the constructor and provide a utility static method to generate an instance with numpy's rng similar to `new_instance`.
:( no, then the model would not necessarily have the properties. It is solvable model if people will put something else in it it would not be that model. (for more info about the SK model please see: https://arxiv.org/abs/1211.1094)
Ultimately we wont this to have these properties.

t-imamichi: As for `new_instance`, I think it's better to make it a static method as we do in other classes. To try many instances the following codes are essentially same.
```python
# current impl
meo = MinimumEigenOptimizer(...)
mod = SKModel(4, np.random.default_rng(1))
for i in range(n):
    qp = mod.to_quadratic_program()
    result = meo.solve(qp)
    ... process results
    mod.new_instance()
``` 

```python
# my proposal
meo = MinimumEigenOptimizer(...)
for i in range(n):
    mod = SKModel.random_instance(4, np.random.default_rng(i))
    qp = mod.to_quadratic_program()
    result = meo.solve(qp)
    ... process results
``` 

To reproduce k-th SK model, the first code needs to iterate k times. But the second one can generate k-th model directly.
```python
# current impl
mod = SKModel(4, np.random.default_rng(1))
for i in range(k-1):
    mod.new_instance()

# my proposal
mod = SKModel.random_instance(4, np.random.default_rng(k))
```

X-Libor: I am not clear why `mod = SKModel(4, np.random.default_rng(k))` in "current implementation"
does not do what you propose?

t-imamichi: Users may want to manually set edge weight either 1 or -1. For example, they may want to flip an edge weight and solve it. If you need to ensure all edge weights are ±1, you can check it in the constructor and raise a warning if not satisfied.

t-imamichi: You make an instance `problem = SKModel(self._num_of_sites, np.random.RandomState(self._seed))` in the unit test. The difference is not essential. I put a random number generator while you use legacy random state.
https://numpy.org/doc/stable/reference/random/generator.html
https://numpy.org/doc/stable/reference/random/legacy.html

t-imamichi: My question is suppose you try  n different `SKModel`s and how do you get the n-1 th model? Is it possible to directly obtain one using `new_instance`?

t-imamichi:
The discussion threads get long. So, let me summarize it.

1. Input of the constructor: This PR receives the number of sites and random number generator as input of the constructor to ensure all edge weights satisfy the SK model property, i.e., ±1. I proposed to make edge weights as input so that users can manually set them. They may give edge weights that do not satisfy the property ±1. I think the constructor can raise a warning in such case.
```python
# PR
def __init__(self, num_of_sites: int, rng: np.random.RandomState = np.random.RandomState(0))

# my proposal
def __init__(self, num_of_sites: int, edge_weights: np.ndarray)
```
2. Destructive update of instance: This PR includes a public method `new_instance` to destructively update the edge weights. I proposed to make a new instance with new edge weights, e.g.,
```python
@staticmethod def random_instance(num_sites, rng: np.random.Generator)
```

Minor point:
- This PR uses the [legacy random state](https://numpy.org/doc/stable/reference/random/legacy.html) of numpy and I propose to use the new [random generator](https://numpy.org/doc/stable/reference/random/generator.html).

If anything missing, feel free to add comments @X-Libor.

qiskit_optimization/applications/sk_model.py
t-imamichi: I maybe got your idea. If you need to reproduce a model you use the constructor; and if you don't have to reproduce any model, you use `new_instance` for efficiency. Is it correct?

X-Libor: > Users may want to manually set edge weight either 1 or -1. For example, they may want to flip an edge weight and solve it. If you need to ensure all edge weights are ±1, you can check it in the constructor and raise a warning if not satisfied.

If they put there their own weights they could do it in a way that does not conform with the model. something which has no mean 0 and variance 1.

X-Libor: > I maybe got your idea. If you need to reproduce a model you generate models with the constructor; and if you don't have to reproduce any model, you use `new_instance` to generate models for efficiency. Is it correct?

I dont understand. if you generate new instances it is reproducible if your pseudo random number generator is independent.

Use case: what is the expected energy  2-layer qaoa achieves for SK model with 20 sites. Does the variational parameters concentrate around some value, and so.

X-Libor: > I follow your unit test `problem = SKModel(self._num_of_sites, np.random.RandomState(self._seed))`. The difference is not essential. I put a random number generator while you use legacy random state. Why do you want to use a legacy one?
> https://numpy.org/doc/stable/reference/random/generator.html
> https://numpy.org/doc/stable/reference/random/legacy.html

https://github.com/Qiskit/qiskit-optimization/pull/208#discussion_r675874389

X-Libor:
> The discussion threads get long. So, let me summarize it to ask for comments of other developers.
> 
>     1. Input of the constructor: This PR receives the number of sites and random number generator as input of the constructor to ensure all edge weights satisfy the SK model property, i.e., ±1. I proposed to make edge weights as input so that users can manually set them. They may give edge weights that do not satisfy the property ±1. I think the constructor can raise a warning in such case.
> 
> 
> ```python
> # PR
> def __init__(self, num_of_sites: int, rng: np.random.RandomState = np.random.RandomState(0))
> 
> # my proposal
> def __init__(self, num_of_sites: int, edge_weights: np.ndarray)
> ```
> 
>     1. Destructive update of instance: This PR includes a public method `new_instance` to destructively update the edge weights to evaluate different edge weights efficiently. I proposed to make a new instance with new edge weights, e.g.,
> 
> 
> ```python
> @staticmethod
> def random_instance(num_sites, rng: np.random.Generator)
> ```
> 
> It might not be as efficient as this PR but I think my proposal is easier to reproduce an instance.
> 
> Minor point:
> 
>     * This PR uses the [legacy random state](https://numpy.org/doc/stable/reference/random/legacy.html) of numpy and I propose to use the new [random generator](https://numpy.org/doc/stable/reference/random/generator.html).
> 
> 
> If anything missing, feel free to add comments @X-Libor.

I cannot comment directly to this. 
Checking just \pm 1 is not sufficient for the asymptotic behavior the model has. It is a solvable model and the weights are random from some distribution that has mean 0 and variance 1. (please see the SK model review I shared).

t-imamichi:
If we evaluate the asymptotic behavior, we need to solve multiple instances of the model. What kind of usage do you have in mind? Could you give me an example of how to evaluate the asymptotic behavior with this class?

As for performance, it would be good to avoid using networkx as an internal data structure because it is slow. I'm actually working on replacing all networkx in applications with retworkx #150 because the latest retworkx supports visualization as networkx.

I think 2D ndarray is a better option because you make a complete graph and 2D ndarray can directly represents weights. If you return weights as ndarray instead of `graph` method, users can generate networkx graph for visualization. It is more efficient because users need to generate an networkx object only when necessary.

```python
import numpy as np
rng = np.random.default_rng(1)

a = np.triu(rng.choice([-1, 1], (4, 4)), k=1)
b = a + a.T # symmetric weights of ±1
print(b)
# [[ 0  1  1  1]
#  [ 1  0  1  1]
#  [ 1  1  0 -1]
#  [ 1  1 -1  0]]

import networkx as nx
g = nx.Graph(b)
print(g.edges.data('weight'))
# [(0, 1, 1), (0, 2, 1), (0, 3, 1), (1, 2, 1), (1, 3, 1), (2, 3, -1)]
```

X-Libor:
> If we evaluate the asymptotic behavior, we need to solve multiple instances of the model. What kind of usage do you have in mind? Could you give me an example of a typical use case of this class?
> 
> As for performance, it would be good to avoid using networkx as an internal data structure because it is slow. I'm actually working on replacing all networkx in applications with retworkx #150 because the latest retworkx supports visualization as networkx.
> 
> I think 2D ndarray is a better option because you make a complete graph and 2D ndarray can directly represents weights. If you return weights as ndarray instead of `graph` method, users can generate networkx graph for visualization. It is more efficient because users need to generate an networkx object only when necessary.
> 
> ```python
> import numpy as np
> rng = np.random.default_rng(1)
> 
> a = np.triu(rng.choice([-1, 1], (4, 4)), k=1)
> b = a + a.T # symmetric weights of ±1
> print(b)
> # [[ 0  1  1  1]
> #  [ 1  0  1  1]
> #  [ 1  1  0 -1]
> #  [ 1  1 -1  0]]
> 
> import networkx as nx
> g = nx.Graph(b)
> print(g.edges.data('weight'))
> # [(0, 1, 1), (0, 2, 1), (0, 3, 1), (1, 2, 1), (1, 3, 1), (2, 3, -1)]
> ```

Can we have a discussion thread for this? I am sorry, I cannot reply to your comments.

*Even if you do not evaluate the asymptotic properties you want the model to actually **be the model** that it claims to be.
*You can always get rid of all graphs and represent them with adjacency/Laplacian matrices. Do you want this? Why is it so important. How slow is it? I consider this to be a scientific and **not a production application** --- people are actually using SK to benchmark QAOA (and so). (BTW: I don't believe that this will be a bottle neck.)

t-imamichi:
I talked with @X-Libor offline. I understood that this application is used like a random instance generator and he does not want users to define instances manually. I personally like to implement an instance generator as a static method, but I'm OK to do it in the constructor.
I checked recirq for reference and found that they implemented the generator as a function and the problem instance as a data class.
https://github.com/quantumlib/ReCirq/blob/144323f212757775a8541dbe2f938b64534b25ec/recirq/qaoa/problems.py#L203

qiskit_optimization/applications/SK_model.py
t-imamichi: I suggest using Random Generator instead of legacy Random State. The `RandomState` documentation says that
> This class should only be used if it is essential to have randoms that are identical to what would have been produced by previous versions of NumPy.

https://numpy.org/doc/stable/reference/random/legacy.html

woodsp-ibm: I agree that Generator based is the right way.

Either way what is the behavior you expect with a default object in the constructor like that. Do you expect it you instantiate two of these just number of sites, of the same value - since they appear to default with the same RandomState(0) that they will produce the same outcome. This is not the way initialization works unfortunately - arguably it might be better to default this to None and then if None internally create the object like that. This way the rng will be per instance as it stands now its shared across instances - which may not be self evident!

X-Libor: Thank you @woodsp-ibm . I made the pseudo rng optional, and changed it to the current one. I hope there are no repeatability issues with it. Can you please check?

X-Libor:
> The discussion threads get long. So, let me summarize it to ask for comments of other developers.
> 
>     1. Input of the constructor: This PR receives the number of sites and random number generator as input of the constructor to ensure all edge weights satisfy the SK model property, i.e., ±1. I proposed to make edge weights as input so that users can manually set them. They may give edge weights that do not satisfy the property ±1. I think the constructor can raise a warning in such case.
> 
> 
> ```python
> # PR
> def __init__(self, num_of_sites: int, rng: np.random.RandomState = np.random.RandomState(0))
> 
> # my proposal
> def __init__(self, num_of_sites: int, edge_weights: np.ndarray)
> ```
> 
>     1. Destructive update of instance: This PR includes a public method `new_instance` to destructively update the edge weights to evaluate different edge weights efficiently. I proposed to make a new instance with new edge weights, e.g.,
> 
> 
> ```python
> @staticmethod
> def random_instance(num_sites, rng: np.random.Generator)
> ```
> 
> It might not be as efficient as this PR but I think my proposal is easier to reproduce an instance.
> 
> Minor point:
> 
>     * This PR uses the [legacy random state](https://numpy.org/doc/stable/reference/random/legacy.html) of numpy and I propose to use the new [random generator](https://numpy.org/doc/stable/reference/random/generator.html).
> 
> 
> If anything missing, feel free to add comments @X-Libor.

@t-imamichi I changed the pseudo rng. Please check.

X-Libor:
> > The discussion threads get long. So, let me summarize it to ask for comments of other developers.
> > ```
> > 1. Input of the constructor: This PR receives the number of sites and random number generator as input of the constructor to ensure all edge weights satisfy the SK model property, i.e., ±1. I proposed to make edge weights as input so that users can manually set them. They may give edge weights that do not satisfy the property ±1. I think the constructor can raise a warning in such case.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```python
> > # PR
> > def __init__(self, num_of_sites: int, rng: np.random.RandomState = np.random.RandomState(0))
> > 
> > # my proposal
> > def __init__(self, num_of_sites: int, edge_weights: np.ndarray)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > 1. Destructive update of instance: This PR includes a public method `new_instance` to destructively update the edge weights to evaluate different edge weights efficiently. I proposed to make a new instance with new edge weights, e.g.,
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```python
> > @staticmethod
> > def random_instance(num_sites, rng: np.random.Generator)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It might not be as efficient as this PR but I think my proposal is easier to reproduce an instance.
> > Minor point:
> > ```
> > * This PR uses the [legacy random state](https://numpy.org/doc/stable/reference/random/legacy.html) of numpy and I propose to use the new [random generator](https://numpy.org/doc/stable/reference/random/generator.html).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If anything missing, feel free to add comments @X-Libor.
> 
> I cannot comment directly to this.
> Checking just \pm 1 is not sufficient for ensuring the asymptotic behavior of the model.
> It is a solvable model and the weights are random from some distribution that has mean 0 and variance 1.
> (please see the SK model review by Panchenko I shared).

Dear @t-imamichi I adressed the legacy pseudo rng. Please see.

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,108 @@
...
+        Constructor for the SK model.
+
+        Args:
+            num_of_sites: number of sites
+            rng: numpy pseudo-random number generator or None.
+            For repeatable experiments use e.g. np.random.default_rng(<SEED>).
+            None results in usage of np.random.
+        """
+        if rng is None:
+            self._rng = np.random
```
t-imamichi, line 52: Because `np.ranom` is a module, `np.random.default_rng()` seems more consistent when `None` is given. You don't need to add  `type: ignore`.
```suggestion
            self._rng = np.random.default_rng()
```

t-imamichi:
Could you add a release note following the [instruction](https://github.com/Qiskit/qiskit-optimization/blob/main/CONTRIBUTING.md#release-notes)?

qiskit_optimization/applications/sk_model.py
X-Libor: thank you, I didnt know how to deal with that. Even asked Dariusz and this was the solution.

X-Libor:
> Could you add a release note following the [instruction](https://github.com/Qiskit/qiskit-optimization/blob/main/CONTRIBUTING.md#release-notes)?

Done. Please see --- it might be "Much Ado About Nothing".

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,108 @@
...
+    where :math:`x_i\in\{\pm 1\}` and :math:`w_{i,j}\in\{\pm 1\}` are chosen independently and uniformly
+    at random. Notice that there are other variants e.g., with :math:`w_{i,j}` chosen from the normal
+    distribution with mean 0 and variance 1.
+
+    References:
+        [1]: Dmitry Panchenko. "The Sherrington-Kirkpatrick model: an overview",
+        https://arxiv.org/abs/1211.1094
+    """
+
+    def __init__(self, num_of_sites: int, rng: Optional[np.random.Generator] = None):
```
woodsp-ibm, line 41: One thing might make it easier for people if this were `Optional[Union[int, np.random.Generator]] = None` They can then simply pass an integer as a seed. They only need to go to a Generator if they need that level of control/definition over the random numbers. Then the code here to set up the self._rng can become something like
```
if isinstance(rng, np.random.Generator):
  self._rng = rng
else 
  self._rng = np.random.default_rng(rng)
```
The latter ie `np.random.default_rng(rng)` will get None or the int seed - where None ends up being seeded randomly from the OS by numpy.

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,108 @@
...
+            self._rng = np.random.default_rng()
+        else:
+            self._rng = rng
+
+        self._num_of_sites = num_of_sites
+        self._graph = nx.complete_graph(self._num_of_sites)
+
+        self.new_disorder()
+
+    def new_disorder(self) -> None:
```
woodsp-ibm, line 61: Would not `disorder()` be enough - in that its going to change things. Do you think it really needs `new_` there?

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def to_quadratic_program(self) -> QuadraticProgram:
+        """Convert an SK model problem instance into a
+        :class:`~qiskit_optimization.problems.QuadraticProgram`.
+
+        Returns:
+            The :class:`~qiskit_optimization.problems.QuadraticProgram` created
+            from the SK problem instance.
+        """
+        mdl = Model(name="SK-model")
```
woodsp-ibm, line 74: Given that the goal seemed to be to generate a bunch of different instances of this class I was wondering with them all having the same name here. On the other hand it may be hard to find a way to make this more unique to each disorder that is simple/easy/short. I guess you can figure things backwards from the objective if you want to get back to the original SK problem given you have a QP that was created from here. Maybe there is no need - but with the one instance changing disorder and emitting a bunch of corresponding QPs was just thinking about a user dealing with that.

releasenotes/notes/sk-model-851851e42be88cc4.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Introduced the Sherrington-Kirkpatrick (SK) model :class:`qiskit_optimization.applications.SKModel`.
```
woodsp-ibm, line 4: So in case you haven't seen these files are collected and form the release notes that end up here https://qiskit.org/documentation/optimization/release_notes.html. I see you had a slightly longer one with an example. I think many users may not know what this is and so at least having a reference link here and perhaps a sentence or two describing what you use this application/model for would be helpful. I saw you had an example - that can work too if you think its informative for a user. How about a tutorial on this? There the example can be runnable along with more description/information to help users understand/use the code etc.

qiskit_optimization/applications/sk_model.py
X-Libor: To be honest I don't know. I would pass that for now. If someone is interested. He/she can add it in another release.

qiskit_optimization/applications/sk_model.py
X-Libor: ok.

qiskit_optimization/applications/sk_model.py
X-Libor: ok.

releasenotes/notes/sk-model-851851e42be88cc4.yaml
X-Libor: Thank you. Is there a way how to locally display the release notes?

X-Libor: Nevertheless, I added there longer text with reference to a review paper. Regarding tutorial, I talked with Stefan and we leave it for another PR.

qiskit_optimization/applications/sk_model.py
woodsp-ibm: I notice in the code you also did an extra check for `int` to separate it away from None. That's not really needed as `np.random.default_rng(None)` is equivalent to doing `np.random.default_rng()` which is why I allowed both int and None to flow through the else part above. In fact I chose to avoid doing an int check since there are various types e.g from numpy that can cause problems with something too specific.

X-Libor: Thank you. Fixed.

releasenotes/notes/sk-model-851851e42be88cc4.yaml
X-Libor: @woodsp-ibm thank you. Please check the release note. If it is ok for you.

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,110 @@
...
+        [1]: Dmitry Panchenko. "The Sherrington-Kirkpatrick model: an overview",
+        https://arxiv.org/abs/1211.1094
+    """
+
+    def __init__(
+        self, num_of_sites: int, rng_or_seed: Optional[Union[np.random.Generator, int]] = None
+    ):
+        """
+        Constructor for the SK model.
+
```
woodsp-ibm, line 46:
```suggestion
```
When the html is built the class docstring and constructor docstring are merged.  We try to avoid having something like this, that states what the code tells you anyway.

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,110 @@
...
+        Args:
+            num_of_sites: number of sites
+            rng_or_seed: numpy pseudo-random number generator or seed for default_rng(.) or None.
+            For repeatable experiments use e.g. np.random.default_rng(<SEED>).
+            None results in usage of np.random.default_rng()
```
woodsp-ibm, line 51:
```suggestion
            rng_or_seed: Optional numpy pseudo-random number generator or seed for default generator. 
                 An internal generator, an instance of numpy random.default_rng(), is used by default; this
                 internal genertaorNone results in usage of np.random.default_rng()
```

woodsp-ibm, line 51:
```suggestion
            rng_or_seed: numpy pseudo-random number generator or seed for default_rng(.) or None.
            For repeatable experiments use e.g. np.random.default_rng(<SEED>).
            None results in usage of np.random.default_rng()
```
If you look at the html generated for this it comes out incorrectly due to lack of indenting I think of the text that is part of the parameter.

image

Can we say something more like the following, that maybe is a little clearer - at least that was my intent. I think the above might be confusing as to whether the user needs to pass np.random.default_rng() with SEED to get reproducible behavior. Well I guess they could since we allow a Generator to be passed but its easier to just pass the seed and the class here builds that out internally.

An optional seed for its default generator, or a NumPy random `Generator` instance to source pseudo-random numbers. By default an np.random.default_rng() generator instance is used internally where this will be created with the seed, if it was passed, to allow experiments to be reproducible.

qiskit_optimization/applications/sk_model.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+
+    def __init__(
+        self, num_of_sites: int, rng_or_seed: Optional[Union[np.random.Generator, int]] = None
+    ):
+        """
+        Constructor for the SK model.
+
+        Args:
+            num_of_sites: number of sites
```
woodsp-ibm, line 48: Do you want to call this 'num_of_sites`? I think more normally people would just use `num_sites`. Also is there anything a little more that can be said here to maybe just say what the sites are that would be helpful for users. Otherwise the descriptive text is not saying anything more that the parameter name tells you itself.

releasenotes/notes/sk-model-851851e42be88cc4.yaml
woodsp-ibm: Release note looks good to me. I did make a couple of other comments though in looking at the documentation that comes out for the class itself.

qiskit_optimization/applications/sk_model.py
X-Libor: If you want to change it I would recommend `no_sites` (https://en.wikipedia.org/wiki/Numero_sign).

qiskit_optimization/applications/sk_model.py
X-Libor: @woodsp-ibm thank you. I changed the text. Sorry I didn't updated it after I added the seed as an option.

qiskit_optimization/applications/sk_model.py
woodsp-ibm: Generally I think Qiskit just uses `num_xxx` e.g. `num_qubits` https://github.com/Qiskit/qiskit-terra/blob/91dfa547acb542d7b374da0c4f600511cd40cf0d/qiskit/circuit/quantumcircuit.py#L1840

qiskit_optimization/applications/sk_model.py
woodsp-ibm: A tiny thing - there does seem to be an extra . character in there in the default_rng method call

> NumPy pseudo-random number generator or seed for np.random.default_rng(.) or None. None results in usage of np.random.default_rng().

i.e. in this this as below
```
np.random.default_rng(.) 
```

qiskit_optimization/applications/sk_model.py
X-Libor: :( but ok.

qiskit_optimization/applications/sk_model.py
X-Libor: Changed to <seed>. BTW, it is a TCS way of denoting that the function has an argument that you use.

woodsp-ibm (approved):
Thanks for making this contribution and the various changes along the way. It looks good to me at this point

t-imamichi (approved):

PR #214, opened 2021-07-29
Allow Qiskit Optimizers in the runtime programs

Cryoris:
### Summary

Follow up on Qiskit/qiskit-nature#290. Allow Qiskit Optimizers objects as ``optimizers`` in the runtime programs instead of only dictionaries.


woodsp-ibm:
Can you add a reno too please

qiskit_optimization/runtime/vqe_program.py
```diff
@@ -309,8 +308,8 @@ class VQEProgramResult(VQEResult):
 
     def __init__(self) -> None:
         super().__init__()
-        self._job_id: Optional[str] = None
-        self._optimizer_history: Optional[Dict[str, Any]] = None
+        self._job_id = None  # type: str
```
t-imamichi, line 311: Why do you need to revert the type hint `Optional[str]` to comment `# type: str`?

woodsp-ibm: I am guessing this was a direct copy of the updated module in Nature that looks like that. I know this module was a little different before - presumably since it was added afterwards and had some changes.

t-imamichi: I see. I prefer type hint personally, but, the code should be consistent with the program in nature. I'm OK with the current code.

Cryoris: Yes, @woodsp-ibm is right I copied the code from nature so that we don't maintain two different codes that do the same (until we find a solution where we only have the code once and not copy+pasted in the first place) 🙂

releasenotes/notes/runtime-optimizers-58f30311f969f54f.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
```
woodsp-ibm, line 2:
```suggestion
fixes:
```

woodsp-ibm (approved):

PR #246, opened 2021-09-03
Rename runtime programs to clients

Cryoris:
### Summary

In the context of Qiskit Runtime, there's been a lot of confusion about what is considered a "runtime program" or "runtime script". To clear this confusion up, we rename the algorithm classes that remotely call the code on the runtime servers to _clients_ and the code on the runtime can be referred to as _program_.

We chose the word client since that's what they are: an API integrated into the algorithm flow that defers the algorithm call to the cloud.

Analogous to Qiskit/qiskit-nature#351.

### Details and comments

The diff might show a lot below, but it was just a renaming of the `VQE(QAOA)Program` to `VQE(QAOA)RuntimeClient` and no internal methods have been touched. 

We could also just call it `VQEClient` for brevity, if that's preferred 🙂

### DIFF
--- a/qiskit_optimization/runtime/qaoa_program.py
+++ b/qiskit_optimization/runtime/qaoa_program.py
@@ -10,26 +10,29 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-"""The Qiskit Optimization QAOA Quantum Program."""
+"""The Qiskit Optimization VQE Quantum Program."""
 
 
 from typing import List, Callable, Optional, Any, Dict, Union
+import warnings
 import numpy as np
 
 from qiskit import QuantumCircuit
-from qiskit.algorithms import MinimumEigensolverResult
-from qiskit.algorithms.optimizers import Optimizer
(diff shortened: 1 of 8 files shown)


t-imamichi:
Will it be backported to stable?

qiskit_optimization/runtime/qaoa_program.py
```diff
@@ -10,26 +10,29 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-"""The Qiskit Optimization QAOA Quantum Program."""
+"""The Qiskit Optimization VQE Quantum Program."""
```
t-imamichi, line 13:
```suggestion
"""The Qiskit Optimization QAOA Quantum Program."""
```

woodsp-ibm:
> Will it be backported to stable?

No. We cannot (should not!!) deprecate code in a bugfix release. Though I have to wonder with the renaming whether the next version of Optimization and Nature should release more or less together otherwise, for users that might be using both, it might seem confusing with the naming.

PR #247, opened 2021-09-03
Improve project setup long-description

manoelmarques:
### Summary
Use readme as project long description in Pypi.


woodsp-ibm:
Since you are editing the readme can I ask if uou can also fix the tutorials link in the readme in the Futher examples section. That is the old one and it should be the new link now i.e. https://qiskit.org/documentation/optimization/tutorials/index.html (not https://qiskit.org/documentation/tutorials/optimization/index.html which is what is there now). I had a note to check/fix these all up - since it was noticed/fixed on one of the other application repos - but since you are fixing this up...

woodsp-ibm (approved):

PR #254, opened 2021-09-22
Re-enable lint f-string check

manoelmarques:

### DIFF
--- a/qiskit_optimization/applications/optimization_application.py
+++ b/qiskit_optimization/applications/optimization_application.py
@@ -54,7 +54,7 @@ class OptimizationApplication(ABC):
         else:
             raise TypeError(
                 "Unsupported format of result. Provide an　OptimizationResult or a",
-                f"binary array using np.ndarray instead of {type(result)}",
+                f" binary array using np.ndarray instead of {type(result)}",
             )
         return x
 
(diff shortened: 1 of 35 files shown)


qiskit_optimization/applications/optimization_application.py
```diff
@@ -54,7 +54,7 @@ def _result_to_x(self, result: Union[OptimizationResult, np.ndarray]) -> np.ndar
         else:
             raise TypeError(
                 "Unsupported format of result. Provide an　OptimizationResult or a",
-                "binary array using np.ndarray instead of {}".format(type(result)),
+                f"binary array using np.ndarray instead of {type(result)}",
```
woodsp-ibm, line 57:
```suggestion
                f" binary array using np.ndarray instead of {type(result)}",
```
I think it ought to have had a space if this and the above string get concatenated otherwise it will read `or abinary` instead of of `or a binary`

qiskit_optimization/applications/tsp.py
```diff
@@ -167,7 +163,7 @@ def parse_tsplib_format(filename: str) -> "Tsp":
                     typ.strip()
                     if typ != "TSP":
                         raise QiskitOptimizationError(
-                            'This supports only "TSP" type. Actual: {}'.format(typ)
+                            f'This supports only "TSP" type. Actual: {typ}'
```
woodsp-ibm, line 166: I thought that via black strings ended up being in double quotes so a bit surprised to see this in single quotes along with the same thing a but further down this file

manoelmarques: Black ignored because the TSP constant already has double-quotes. It doesn't touch strings like that. There are other cases.

woodsp-ibm (approved):

PR #257, opened 2021-09-25
Duck-type VQE interim results

renier:
### Summary

Add a check for the type of data expected in the callback wraper in VQEProgram.

This will allow us to close Qiskit-Partners/qiskit-runtime#38

### Details and comments

Anticipating changes in the Qiskit Runtime API which will start emitting final results in the WebSocket stream. When that happens, if we don't duck-type in the VQE callback wrapper receiving the interim results, it will get a ValueError exception.

Already done in Qiskit/qiskit-nature#350

### DIFF
--- a/qiskit_optimization/runtime/vqe_program.py
+++ b/qiskit_optimization/runtime/vqe_program.py
@@ -197,7 +197,7 @@ class VQEProgram(MinimumEigensolver):
 
     @property
     def callback(self) -> Callable[[int, np.ndarray, float, float], None]:
-        """Returns the call:back."""
+        """Returns the callback."""
         return self._callback
 
     @callback.setter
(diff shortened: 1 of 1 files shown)


qiskit_optimization/runtime/vqe_program.py
```diff
@@ -196,25 +196,28 @@ def store_intermediate(self, store: bool) -> None:
         self._store_intermediate = store
 
     @property
-    def callback(self) -> Callable:
-        """Returns the callback."""
+    def callback(self) -> Callable[[int, np.ndarray, float, float], None]:
+        """Returns the call:back."""
```
woodsp-ibm, line 200:
```suggestion
        """Returns the callback."""
```

woodsp-ibm (approved):

t-imamichi (approved):

PR #258, opened 2021-09-27
Create bin_packing.py

alejomonbar:
Adding the bin packing problem to the applications on qiskit_optimization

### Summary

I add the [bin packing problem](https://en.wikipedia.org/wiki/Bin_packing_problem) to the set of applications of qiskit optimization.

### Details and comments
**Example**

```python
from qiskit_optimization.applications import BinPacking
import numpy as np
from qiskit_optimization.algorithms import CplexOptimizer

np.random.seed(1)

num_items = 3
max_number_of_bins = num_items
max_weight = 15
weights = [np.random.randint(max_weight) for i in range(num_items)]

bin_packing = BinPacking(weights, max_weight, max_number_of_bins)
qp = bin_packing.to_quadratic_program()

optimizer = CplexOptimizer() # Optimizer
result = optimizer.solve(qp) #Solving the bin_packing problem created

bin_in_packings = bin_packing.interpret(result) # Converting the bin packing problem 
# in something we can interpret.Each list in the bin_in_packings list is a bin and each
# element in that list is an item.

fig = bin_packing.draw(result) #Create a plot of the results in the bins
fig.show() #The results are store as a plt.Figure
fig.savefig("items_in_bins.png") # We can save the results 
```
items_in_bins


qiskit_optimization/applications/bin_packing.py
```diff
@@ -0,0 +1,97 @@
...
+
+class BinPacking(OptimizationApplication):
+    """Optimization application for the "bin packing" [1] problem.
+
+    References:
+        [1]: "Bin packing",
+        https://en.wikipedia.org/wiki/Bin_packing_problem
+    """
+
+    def __init__(self, weights: List[int], max_weight: int) -> None:
```
t-imamichi, line 34: It might be good to have `max_num_bins` as `Optional[int]`. If it is `None`, you can set the number of items as `max_num_bins` as the trivial upper bound.

qiskit_optimization/applications/bin_packing.py
```diff
@@ -0,0 +1,97 @@
...
+    @property
+    def max_weight(self) -> int:
+        """Getter of max_weight
+
+        Returns:
+            The maximal weight for the bin packing problem
+        """
+        return self._max_weight
+
+    @max_weight.setter
```
t-imamichi, line 90: I would suggest to remove setter even if other apps have one. Users can make another instance by creating a new instance. If you compute the non-trivial upper bound of number of bins (such as first fit), setter makes the code complicated.

qiskit_optimization/applications/bin_packing.py
```diff
@@ -0,0 +1,97 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2018, 2021.
```
woodsp-ibm, line 3: Minor comment - since this is new code then just the 2021 date is all that is needed.

qiskit_optimization/applications/bin_packing.py
```diff
@@ -0,0 +1,97 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""An application class for the set packing."""
```
woodsp-ibm, line 13: This comment comes from the class upon which this was based I guess and needs changing from set packing

qiskit_optimization/applications/bin_packing.py
alejomonbar: Fix it.

qiskit_optimization/applications/bin_packing.py
alejomonbar: Fix it.

qiskit_optimization/applications/bin_packing.py
```diff
@@ -31,14 +31,19 @@ class BinPacking(OptimizationApplication):
...
 
-    def __init__(self, weights: List[int], max_weight: int) -> None:
+    def __init__(
+        self, weights: List[int], max_weight: int, max_number_of_bins: Optional[int] = None
+    ) -> None:
         """
         Args:
             weights: A list of the weights of items
             max_weight: The maximum bin weight capacity
+            max_number_of_bins: The maximum number of bins
```
woodsp-ibm, line 41: Can we add here that by default this is set to the number of weights to make it clear in the documentation.

test/applications/test_bin_packing.py
```diff
@@ -0,0 +1,75 @@
...
+"""Test bin packing class"""
+from test.optimization_test_case import QiskitOptimizationTestCase
+
+from qiskit_optimization import QuadraticProgram
+from qiskit_optimization.algorithms import OptimizationResult, OptimizationResultStatus
+from qiskit_optimization.applications.bin_packing import BinPacking
+from qiskit_optimization.problems import Constraint, QuadraticObjective, VarType
+
+
+class TestBinPacking(QiskitOptimizationTestCase):
```
woodsp-ibm, line 22: Can a test be added that covers max_number_of_bins i.e a non-default value for this.

qiskit_optimization/applications/bin_packing.py
```diff
@@ -31,14 +31,19 @@ class BinPacking(OptimizationApplication):
...
         """
         self._weights = weights
         self._max_weight = max_weight
+        if max_number_of_bins is None:
+            self.max_number_of_bins = len(weights)
```
woodsp-ibm, line 46: Should this be self._max_number_of_bins so its marked as private variable like the others (I don't think we want to expose this as part of the public API right - which as it stands, not being marked private, it appears to be)

test/applications/test_bin_packing.py
```diff
@@ -0,0 +1,75 @@
...
+    def test_interpret(self):
+        """Test interpret"""
+        bin_packing = BinPacking(weights=self.weights, max_weight=self.max_weight)
+        self.assertEqual(bin_packing.interpret(self.result), [0, 2, 3, 4, 11])
+
+    def test_max_weight(self):
+        """Test max_weight"""
+        bin_packing = BinPacking(weights=self.weights, max_weight=self.max_weight)
+        bin_packing.max_weight = 5
+        self.assertEqual(bin_packing.max_weight, 5)
```
woodsp-ibm, line 75: Isn't this test now kinda bogus? There is no max_weight property anymore - since Python allows you to add fields to any object then line 74 is just adding a new field max_weight with value 5 and the next line reads it back and passes.

test/applications/test_bin_packing.py
```diff
@@ -0,0 +1,75 @@
...
+        self.assertEqual(lin[0].rhs, 1)
+        self.assertEqual(lin[0].linear.to_dict(), {3: 1.0, 6: 1.0, 9: 1.0})
+        self.assertEqual(lin[3].sense, Constraint.Sense.LE)
+        self.assertEqual(lin[3].rhs, 0)
+        self.assertEqual(lin[3].linear.to_dict(), {3: 16.0, 4: 9.0, 5: 23.0, 0: -40.0})
+
+    def test_interpret(self):
+        """Test interpret"""
+        bin_packing = BinPacking(weights=self.weights, max_weight=self.max_weight)
+        self.assertEqual(bin_packing.interpret(self.result), [0, 2, 3, 4, 11])
```
woodsp-ibm, line 69: Just a question - I am trying to understand the result here and what it means. This is interpreting some known OptimizationResult - I guess I am wondering how do I get to the bins and weights that are binned therein from what seems like a flat list here.

a-matsuo: I agree with @woodsp-ibm. `interpret` should return some "user-friendly" expression of the solution. In this case, `[ [0, 1], [], [2]]` looks a nice format.

test/applications/test_bin_packing.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def test_interpret(self):
+        """Test interpret"""
+        bin_packing = BinPacking(weights=self.weights, max_weight=self.max_weight)
+        self.assertEqual(bin_packing.interpret(self.result), [[0, 1], [], [2]])
+
+    def test_max_weight(self):
+        """Test max_weight"""
+        bin_packing = BinPacking(weights=self.weights, max_weight=self.max_weight)
+        self.assertEqual(bin_packing._max_weight, 40)
```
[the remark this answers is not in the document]
alejomonbar: Thank you for your comments, I'm learning a lot @woodsp-ibm!

qiskit_optimization/applications/bin_packing.py
```diff
@@ -0,0 +1,93 @@
...
+        x = mdl.binary_var_list([f"x{i}" for i in range(num_bins)])
+        mdl.minimize(mdl.sum([x[i] for i in range(num_bins)]))
+        y = mdl.binary_var_matrix(num_bins, num_items, name="y")
+        for i in range(num_items):
+            # First set of constraints: the items must be in any bin
+            mdl.add_constraint(mdl.sum([y[(j, i)] for j in range(num_bins)]) == 1)
+        for i in range(num_bins):
+            # Second set of constraints: weight constraints
+            mdl.add_constraint(
+                mdl.sum([self._weights[j] * y[(i, j)] for j in range(num_items)])
```
t-imamichi, line 70:
```suggestion
                mdl.sum([self._weights[j] * y[i, j] for j in range(num_items)])
```

qiskit_optimization/applications/bin_packing.py
```diff
@@ -0,0 +1,93 @@
...
+        :class:`~qiskit_optimization.problems.QuadraticProgram`
+
+        Returns:
+            The :class:`~qiskit_optimization.problems.QuadraticProgram` created
+            from the bin packing problem instance.
+        """
+        mdl = Model(name="BinPacking")
+        num_bins = self._max_number_of_bins
+        num_items = len(self._weights)
+        x = mdl.binary_var_list([f"x{i}" for i in range(num_bins)])
```
t-imamichi, line 61:
```suggestion
        x = mdl.binary_var_list(num_bins, name="x")
```

t-imamichi, line 61: Could you revise the model to be consistent with the model of Wikipedia, which you mentioned as a reference?

qiskit_optimization/applications/bin_packing.py
```diff
@@ -0,0 +1,93 @@
...
+from qiskit_optimization.translators import from_docplex_mp
+from .optimization_application import OptimizationApplication
+
+
+class BinPacking(OptimizationApplication):
+    """Optimization application for the "bin packing" [1] problem.
+
+    References:
+        [1]: "Bin packing",
+        https://en.wikipedia.org/wiki/Bin_packing_problem
```
t-imamichi, line 31:
```suggestion
        `https://en.wikipedia.org/wiki/Bin_packing_problem
        <https://en.wikipedia.org/wiki/Bin_packing_problem>`_
```

t-imamichi:
Could you resolve the conflict?

alejomonbar:
> Could you resolve the conflict?

I'm not sure how to solve it. What should I do? 

> Could you resolve the conflict?

Yes, I did and added an extra feature for the visualization @t-imamichi

qiskit_optimization/applications/__init__.py
```diff
@@ -63,7 +65,8 @@
 from .vertex_cover import VertexCover
 from .sk_model import SKModel
 
-__all__ = [
+_all__ = [
```
t-imamichi, line 68: You should not overwrote the fix.
```suggestion
__all__ = [
```

alejomonbar: thank you!

qiskit_optimization/applications/bin_packing.py
```diff
@@ -0,0 +1,149 @@
...
+        num_items = len(self._weights)
+        num_bins = self._max_number_of_bins
+        bins = x[:num_bins]
+        items = np.array(x[num_bins:]).reshape((num_items, num_bins))
+        items_in_bins = [
+            [i for i in range(num_items) if bins[j] and items[i, j]] for j in range(num_bins)
+        ]
+        return items_in_bins
+
+    def draw(self, result: Union[OptimizationResult, np.ndarray]):
```
woodsp-ibm, line 104: This does not seem like the same as the `draw` methods in the graph applications that internally use nx.draw and actually draw the graph. This here seems more like a get_figure/get_plot type method since it returns something that you can then draw. 

Also as it stands it has no return typehint. Before it returned `plt.Figure` but I get that this does not work when MatPlotLib is not installed. Rather perhaps what can be done, assuming what is wanted here is to return a figure that the user can draw, rather than actually drawing it. is to do similar to what is done in TorchConnector in MachineLearing that references types that depend on Torch being installed and would otherwise fail if it is not. Basically this here could return Figure - that would be imported from matplotlib.plt but if that failed then it could just define a dummy Figure locally to satisfy the code. Since it will check whether matplotlib is installed this dummy class is simply to allow the code to get that far, it would not be used in practice.

t-imamichi: FYI: We use a fake class for a similar situation of Gurobi. https://github.com/Qiskit/qiskit-optimization/blob/d911c452b7e82bb9b48ed9655dc8fe9b01bfe6fa/qiskit_optimization/translators/gurobipy.py#L25

alejomonbar: Thank you @woodsp-ibm, I followed your suggestions.

qiskit_optimization/applications/bin_packing.py
```diff
@@ -0,0 +1,157 @@
...
+    import matplotlib.pyplot as plt
+    from matplotlib.pyplot import Figure
+
+    _HAS_MATPLOTLIB = True
+except ImportError:
+    _HAS_MATPLOTLIB = False
+
+    class Figure:  # type: ignore
+        """Empty Figure class
+        Replacement if matplotlib.pyplot.Figure is not present.
```
woodsp-ibm, line 36:
```suggestion
        Replacement Figure for when matplotlib is not present.
```
CI failed on pyplot - we could add pyplot to the custom dictionary - however matplotlib is already there, so maybe this rewording suggestion is fine.

woodsp-ibm:
@t-imamichi CI is now failing around gurobipy - which appears to have had a new release (about 4hrs ago according to pypi at the time I am making this comment). I guess that needs looking at fixing in a different PR to address this. I see there is no new package for 3.6 in the new 9.5.0 release so it ran against the older one, which can explain why it passed. But 3.9 looks like its with the new 9.5.0.

t-imamichi:
@woodsp-ibm OK. I take care of gurobi stuff.

woodsp-ibm:
I see we do not have any unit testing that includes the `get_figure` method. Any unit test would have to skip when matplotlib is not installed but when installed can we do some sort of test? At least make sure it returns a matplotlib figure without failing - testing the figure itself. i.e. the returned object, other than its type is correct may not be realistically possible but maybe you have some ideas.

alejomonbar:
> I see we do not have any unit testing that includes the `get_figure` method. Any unit test would have to skip when matplotlib is not installed but when installed can we do some sort of test? At least make sure it returns a matplotlib figure without failing - testing the figure itself. i.e. the returned object, other than its type is correct may not be realistically possible but maybe you have some ideas.

Hi @woodsp-ibm, I added the test of the figure, let me know what you think.

test/applications/test_bin_packing.py
```diff
@@ -89,3 +96,13 @@ def test_max_number_of_bins(self):
...
+
+    def test_figure(self):
+        """Test the plot of the Bin Packing Problem is properly generated."""
+        bin_packing = BinPacking(
+            weights=self.weights,
+            max_weight=self.max_weight,
+            max_number_of_bins=self.max_number_of_bins,
+        )
+        if _HAS_MATPLOTLIB:
+            self.assertEqual(bin_packing.get_figure(self.result), Figure)
```
t-imamichi, line 108: `assertEqual` is not appropriate. Please use `assertIsInstance`.
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsInstance

test/applications/test_bin_packing.py
```diff
@@ -89,3 +96,13 @@ def test_max_number_of_bins(self):
...
         self.assertEqual(lin[4].linear.to_dict(), {3: 16.0, 5: 9.0, 7: 23.0, 1: -40.0})
+
+    def test_figure(self):
+        """Test the plot of the Bin Packing Problem is properly generated."""
+        bin_packing = BinPacking(
+            weights=self.weights,
+            max_weight=self.max_weight,
+            max_number_of_bins=self.max_number_of_bins,
+        )
+        if _HAS_MATPLOTLIB:
```
t-imamichi, line 107: Please make `else` pattern with assertRaises.
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

alejomonbar: Thank you. I made the changes you suggested.

t-imamichi: Please check the document and fix the test accordingly?
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

manoelmarques: I fixed this test to conform to the way we handle extra libraries not installed in unit test (using decorator).

t-imamichi:
Could you make a unit test where `max_number_of_bins` is not given to `BinPacking`? You need to verify the actual max number of bins determined internally and the correctness of the solution.

alejomonbar:
> Could you make a unit test where `max_number_of_bins` is not given to `BinPacking`? You need to verify the actual max number of bins determined internally and the correctness of the solution.

@t-imamichi This is already tested in `test_to_quadratic_program`, let me know if this is enough or I should do another test.

t-imamichi:
Sorry. I missed it. It looks good.

t-imamichi (approved):
Thanks!

PR #278, opened 2021-11-17
Fix lists in tutorials

t-imamichi:
### Summary

Fix lists in an inappropriate format.

current
image

this PR
image


woodsp-ibm:
That's quite a few places! Sphinx as you have seen needs the extra blank line whereas Jupyter notebook does not when it renders.

woodsp-ibm (approved):

t-imamichi:
Yeah. The problem is Jupyter notebook works fine even if a blank line is missing. It would be nice if there is an option of sphinx that can detect this issue. Or, we perhaps need to make a tool to detect it.

PR #284, opened 2021-11-26
Fix latest pylint errors and mergify automerge

manoelmarques:

### DIFF
--- a/qiskit_optimization/problems/substitute_variables.py
+++ b/qiskit_optimization/problems/substitute_variables.py
@@ -154,16 +154,14 @@ class _SubstituteVariables:
         """Checks feasibility of the following condition
         0 `sense` rhs
         """
-        # I use the following pylint option because `rhs` should come to right
-        # pylint: disable=misplaced-comparison-constant
         if sense == ConstraintSense.EQ:
-            if 0 == rhs:
+            if rhs == 0:
                 return True
         elif sense == ConstraintSense.LE:
-            if 0 <= rhs:
+            if rhs >= 0:
                 return True
         elif sense == ConstraintSense.GE:
-            if 0 >= rhs:
+            if rhs <= 0:
                 return True
         return False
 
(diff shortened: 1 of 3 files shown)


qiskit_optimization/problems/substitute_variables.py
```diff
@@ -154,16 +154,14 @@ def _feasible(sense: ConstraintSense, rhs: float) -> bool:
...
         """
-        # I use the following pylint option because `rhs` should come to right
-        # pylint: disable=misplaced-comparison-constant
         if sense == ConstraintSense.EQ:
-            if 0 == rhs:
+            if rhs == 0:
                 return True
         elif sense == ConstraintSense.LE:
-            if 0 <= rhs:
+            if rhs >= 0:
```
t-imamichi, line 161: I intentionally implemented the conventional code so that the sense (= LE) is consistent with the if clause `0 <= rhs`. If we flip it, they are not consistent. For example, if `sense == ConstraintSense.LE`, we check `rhs >= 0` (>= is GE). I prefer `pylint: disable=misplaced-comparison-constant`.

qiskit_optimization/problems/substitute_variables.py
```diff
@@ -154,16 +154,14 @@ def _feasible(sense: ConstraintSense, rhs: float) -> bool:
...
                 return True
         elif sense == ConstraintSense.GE:
-            if 0 >= rhs:
+            if rhs <= 0:
```
t-imamichi, line 164: ditto

t-imamichi:
I see. `misplaced-comparison-constant` of pylint requires an extension now.

> Moved misplaced-comparison-constant to its own extension comparison_placement.
This checker was opinionated and now no longer a default. It can be reactived by adding
pylint.extensions.comparison_placement to load-plugins in your config.
https://github.com/PyCQA/pylint/releases

qiskit_optimization/problems/substitute_variables.py
t-imamichi: I see `misplaced-comparison-constant` has been moved to an extension of pylint. It does not make sense to activate it only for this case.

t-imamichi (approved):

PR #295, opened 2021-12-09
Update garrison's name in .mailmap

garrison:
### Summary

With this change, my name in `Qiskit.bib` will (eventually) match the name I use for academic publications.


garrison:
Sorry -- actually, this might not be necessary.  I was under the impression that each application had its own `bib` file, but it looks like I was wrong.

The README here says

> Please note that if you do not like the way your name is cited in the BibTex file then consult the information found in the [.mailmap](https://github.com/Qiskit/qiskit-optimization/blob/main/.mailmap) file.

... but it's a little unclear to me whether I need to update this in the `.mailmap` for each module or if doing it only in `qiskit/.mailmap` is enough.

mtreinish:
If the real name in the git log for your commits (or any commits you're listed as a co-author on) for a tracked repository you'll want to ensure the mailmap corrects that to your preferred name. The authors list (and the bibtex file by extension) in `Qiskit/qiskit` is generated by parsing the git log from these repositories https://github.com/Qiskit/qiskit/blob/master/tools/generate_authors.py#L39-L44. So you probably don't strictly need to this here because the qiskit bib file doesn't include optimization.

That being said the script attempts to de-duplicate entries where it can but if the real name or email of a commit don't match it won't be able to so it's generally better to ensure you've adjusted the mailmap to ensure all your preferred name is used consistently everywhere.

mtreinish (approved):

manoelmarques (approved):

Issue #255, opened 2021-09-23
Add Max-k-cut appication


clausia:
### What is the expected behavior?
Being able to solve a Max-k-cut type problem

> In the Max-k-Cut problem, we are given an undirected graph, G=(V, E), with non-negative edge weights. Our objective is to divide the vertices into at most k disjoint sets, for some given positive integer k, to maximize the weight of the edges whose endpoints lie in different sets. When k=2, this problem is known simply as the Max-Cut problem.

----

**Add Max-k-cut application**

We propose to add the functionality to be able to solve the Max-k-cut problem, similar to how the class that contains the Maxcut application already exists.

We have already made progress on this project between Vyron Vasileiadis (@fedonman) and myself (Claudia Zendejas-Morales, @clausia), we raise it in this issue to be able to discuss any concerns that may arise in this regard.

Furthermore, we have analyzed the solution for Maxcut that exists here in this qiskit-optimization project, as well as the solution proposed in work [1].

Please let us know any comments you have on this proposal.

[1] Zsolt Tabi; Kareem H. El-Safty; Zsófia Kallus; Péter Hága; Tamás Kozsik; Adam Glos; Zoltán Zimborás
_Quantum Optimization for the Graph Coloring Problem with Space-Efficient Embedding_
https://ieeexplore.ieee.org/document/9259934

PR #299, opened 2021-12-13
Adding the Max-k-Cut application

clausia:
### Summary

Adding the Max-k-Cut to the application section

### Details and comments

Max-k-Cut: Given an undirected graph, find a partition of nodes into at most k subsets such that the total weight of the edges between the k subsets is maximized. To solve this problem, the space-efficient quantum optimization representation (or encoding) for the graph coloring problem proposed in [1] is used.

This work has been done in collaboration between Claudia Zendejas-Morales (@clausia) and Vyron Vasileiadis (@fedonman). This contribution was done under QIntern 2021 (https://qworld.net/qintern-2021/) organized by QWorld, under the supervision of Adam Glos (@adamglos92) and Zoltan Zimboras (@zimboras). We would like to thank Kareem H. El-Safty (@kareem1925) for bringing up the idea of implementing space-efficient Max-k-Cut on qiskit.

---

[1]: Z. Tabi et al., "Quantum Optimization for the Graph Coloring Problem with Space-Efficient Embedding," 2020 IEEE International Conference on Quantum Computing and Engineering (QCE), 2020, pp. 56-62, doi: 10.1109/QCE49297.2020.00018.

---

Closes https://github.com/Qiskit/qiskit-optimization/issues/255


qiskit_optimization/applications/max_k_cut.py
```diff
@@ -0,0 +1,198 @@
...
+            graph: A graph representing a problem. It can be specified directly as a
+                `NetworkX <https://networkx.org/>`_ graph,
+                or as an array or list format suitable to build out a NetworkX graph.
+            k: The number of colors
+            colors: List of strings or list of colors in rgba lists to be assigned to each
+                resulting subset, there must be as many colors as the number k
+        """
+        super().__init__(graph=graph)
+        self._subsets_num = k
+        self._colors = colors if colors and len(colors) >= k else None
```
woodsp-ibm, line 65: For colors I see that more are allowed to be defined and presumably ignored; however if colors were given but less than k then they are discarded. In looking further, from what can see, these are just used if/when the user the draws the result.

I am wondering whether it might be better to move this off the constructor and just say have a method call something like `set_draw_colors` that takes the same thing and stores it for the rendering. That makes this feel less like part of the problem definition - at least to me. I would have said pass them on draw but then this is implementing logic defined in the base class and it does not seem to support customizing the rendering in this way.

In terms of the colors I think it should define what the string and int formats are - if its simply matplot color format then I think that can be stated perhaps with a link to matplot colot doc that has the info plus perhaps an example in the code to help people.

Also I wonder since this is a set of colors should they be passed in as a set so that it contains unique items - otherwise a user can specify the same color again. In regards of the behavior or more or less colors given than k maybe you prefer the less severe mechanism of discarding all or part of them depending on size. I think at least the code should emit a warning to alert the user - ideally they should be passing in k colors right not some other sized set. If this is a separate method just to set draw colors then maybe allowing only None (to reset to internal default) or a k sized set and raising a ValueError if the size if not equal to k might be more acceptable?

qiskit_optimization/applications/max_k_cut.py
```diff
@@ -0,0 +1,198 @@
...
+        nx.draw(self._graph, node_color=self._node_color(x), pos=pos, with_labels=True)
+
+    def _node_color(
+        self, x: np.ndarray
+    ) -> Union[List[Tuple[float, float, float, float]], List[str]]:
+        # Return a list of colors for draw.
+
+        n = self._graph.number_of_nodes()
+
+        # k colors chosen (randomly or from cm.rainbow), or from given color list
```
woodsp-ibm, line 135: As far as doing things randomly there exists an random generator in algorithm_globals in Terra that we created to have a common generator that algorithms and applications would use so that a user could create reproducible results by seeding it. Using that, instead of whats done here, would allow a user to seed it and have the same colored drawing each time if they prefer - of course they could pass in their own colors too, but I think using the common generator is preferable.

qiskit_optimization/applications/max_k_cut.py
clausia: Thank you for your comments, they have already been addressed:

- The importance of including the definition of colors in the constructor is that this implementation is based on the "Graph Coloring Problem", but I understand your concern, we can have the solution of the max-k-cut problem without the need to draw the resulting graph. To change (or add) the colors you have the methods set/get (`colors`, called like that for the same reason, that we are thinking about the "Graph Coloring Problem").
So, I have removed the `colors` parameter from the constructor, (before it was optional), I have already done this update.

- As for the documentation, certainly including an example is much better for people who use this class, that update is ready, please review.

- Regarding that the user can specify a repeated color, it seems to me that we should allow this type of flexibility, that this decision is up to the user, finally if it was not his/her intention, she/he will be able to see the result when he requests the drawing of the graph (the user will see fewer colors than '`k`').

- Now it is being validated that the size of the color list is equal to the k parameter, and if it is not, the `ValueError` is thrown indicating that the colors have not been assigned because they are not the correct amount.

qiskit_optimization/applications/max_k_cut.py
clausia: Now `algorithm_globals` is used

qiskit_optimization/applications/max_k_cut.py
clausia: @woodsp-ibm, what do you think of the changes?

qiskit_optimization/applications/max_k_cut.py
```diff
@@ -0,0 +1,247 @@
...
+
+    _HAS_MATPLOTLIB = False
+
+
+class Maxkcut(GraphOptimizationApplication):
+    """Optimization application for the "max-k-cut" [1] problem based on a NetworkX graph.
+
+    References:
+        [1]: Z. Tabi et al.,
+             "Quantum Optimization for the Graph Coloring Problem with Space-Efficient Embedding"
```
t-imamichi, line 42: This paper addresses a graph coloring problem and does not mention max-k-cut problem.
Could you refer an appropriate paper or page?

qiskit_optimization/applications/max_k_cut.py
```diff
@@ -0,0 +1,247 @@
...
+            from the Max-k-cut problem instance.
+        """
+        for w, v in self._graph.edges:
+            self._graph.edges[w, v].setdefault("weight", 1)
+
+        mdl = Model(name="Max-k-cut")
+        n = self._graph.number_of_nodes()
+        k = self._subsets_num
+        x = {(v, i): mdl.binary_var(name=f"x_{v}_{i}") for v in range(n) for i in range(k)}
+        first_penalty = mdl.sum_squares((1 - mdl.sum(x[v, i] for i in range(k)) for v in range(n)))
```
t-imamichi, line 80: Could you make a model with a standard formulation of the problem? Please do not penalize constraints by yourself.
In the application classes, we make a model with a standard formulation. The conversion to QUBO will be done within algorithm classes.

qiskit_optimization/applications/max_k_cut.py
```diff
@@ -0,0 +1,247 @@
...
+        second_penalty = mdl.sum(
+            mdl.sum(self._graph.edges[v, w]["weight"] * x[v, i] * x[w, i] for i in range(k))
```
t-imamichi, line 82: I think this takes sum of edges whose end nodes are same color (`i`). Is it a correct k-cut value?
I think k-cut value is the sum of edges whose end nodes have different colors.
Anyways, could you refer an appropriate paper or page and make a model (obj func + constraints).

qiskit_optimization/applications/max_k_cut.py
adamglos92: Hi @t-imamichi, graph colouring problem is a decision variant of max-k-cut. Graph coloring problem is about checking if there "exists colouring of the nodes which make every edge connecting nodes with a different colour." Contrary, Max-K-Cut is about finding the colouring for which the number of such edges is maximized. You could say that graph colouring to max-K-Cut shows similar relation as SAT to MAX-SAT. For this reason, and the fact max cut is already used in the package we decided to choose max-k-cut name

t-imamichi: As far as the paper mentions only coloring problem, I don't think it's an appropriate citation. Why not rename the class "k-coloring problem" as the paper addresses?

t-imamichi: Moreover, there are many papers directly addressing max-k-cut problem. Why do you need to refer that paper (Tabi et al.) for max-k-cut class?

adamglos92: The notion of feasibility changes. For Max-Cut every solution is feasible, for graph colouring only those with all edges connecting different nodes are feasible. This is obviously correct but inconsistent with max-cut used in the qiskit-optimization. 
Since this is for optimization purposes mostly we found Max-K-Cut more accurate. But we can also move it to Max-K-Cut
We are not aware of older papers that propose the given formulation. If there is one, we're open to replacing the reference.

t-imamichi: You may perhaps try to use an solution of coloring problem as a heuristic solution of max-k-cut problem?
If so, I ask you to make a proper mixed integer programming model of max-k-cut problem.
The model of application classes should be exact. Otherwise, even if we apply CPLEX or Gurobi to the model, we may not be able to obtain the optimal solution.

t-imamichi:
@clausia Do you have any update? If no, we will close this PR.

clausia:
@t-imamichi Hi, thanks for asking, I think it's best to close this PR and open another one when the update is ready

PR #304, opened 2022-01-04
Remove unnecessary Union with PauliSumOp

garrison:
### Summary

This replaces `Union[OperatorBase, PauliSumOp]` with `OperatorBase`, as `PauliSumOp` is a subclass of `OperatorBase`, so the union is redundant.

### Details and comments

I'm proposing this change because I found the union misleading -- I expected that its presence implied that `PauliSumOp` is not part of the `OperatorBase` hierarchy, but that is wrong.  I believe this change makes more sense, but if there is a reason to prefer that `PauliSumOp` be listed explicitly in the `Union`, I am open to learning why.


t-imamichi (approved):
Thanks!

PR #308, opened 2022-01-11
Update docplex url

garrison:
### Summary

The previous URL redirects to the https://developer.ibm.com/ front page, so I updated it to be the same URL that is listed in the qiskit-optimization README.

### DIFF
--- a/qiskit_optimization/__init__.py
+++ b/qiskit_optimization/__init__.py
@@ -22,7 +22,7 @@ to a suite of easy-to-use quantum optimization algorithms that are ready to run 
 classical simulators, as well as on real quantum devices via Qiskit.
 
 This module enables easy, efficient modeling of optimization problems using `docplex
-<http://ibmdecisionoptimization.github.io/docplex-doc/>`_.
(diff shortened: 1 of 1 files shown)


qiskit_optimization/__init__.py
```diff
@@ -22,7 +22,7 @@
 classical simulators, as well as on real quantum devices via Qiskit.
 
 This module enables easy, efficient modeling of optimization problems using `docplex
-<https://developer.ibm.com/docloud/documentation/optimization-modeling/modeling-for-python/>`_.
+<http://ibmdecisionoptimization.github.io/docplex-doc/>`_.
```
woodsp-ibm, line 25:
```suggestion
<https://ibmdecisionoptimization.github.io/docplex-doc/>`_.
```
The link in the readme is to https which I think is preferred nowadays over http, so I think that should be here too

garrison: Thanks for catching -- somehow this escaped my notice.

woodsp-ibm (approved):

t-imamichi (approved):

Issue #84, opened 2021-04-15
The tutorial of converters needs to explain how solutions are interpreted


t-imamichi:
### What is the expected enhancement?

The tutorial of converters explains how we apply `convert` only.
But, it is also important how we interpret solutions from the converted problem to the original problem.
Moreover, we need to explain that users don't need to use converters manually in most cases and the algorithms apply converters properly.
 
https://qiskit.org/documentation/optimization/tutorials/02_converters_for_quadratic_programs.html

garrison:
Somewhat related, it would also be very nice if the documentation explained any special considerations to consider when choosing the order in which to apply multiple converters.  For instance, in the constructor to `QuadraticProgramToQubo`:

https://github.com/Qiskit/qiskit-optimization/blob/d3bcf30941010068dc89031bb0a2cd5bc1835b7c/qiskit_optimization/converters/quadratic_program_to_qubo.py#L48-L56

Is there any particular reason that `IntegerToBinary` is before `LinearEqualityToPenalty` yet is after `LinearInequalityToPenalty` and `InequalityToEquality`?

(Ideally, there should also be a comment at that location in the code to explain why the above order makes sense there.)

gprs1809:
Hello @t-imamichi, @garrison, is this issue still open? Do you think I can help with this? I am new to open source contributions, I have only raised 1 pull request so far. I found this to be an interesting issue since I would myself like to understand the topics you have mentioned. Once I understand, I can document these, but let me know if it's too soon to try this, given my less experience with open source. If I can take this up and then, if I get stuck in understanding something, I hope I can discuss it here.

t-imamichi:
@gprs1809 Thank you for your interest in this project. Are you familiar with Qiskit optimization? This issue requires the knowledge of the internals of Qiskit optimization to write details in the tutorial.

t-imamichi:
@garrison Sorry to forget to reply to your question. `IntegerToBinary` should come after all integer variables are defined. Since `InequalityToEquality` introduces integer slack variables, `InequalityToEquality` should be called before `IntegerToBinary`.
I agree that we should write such conditions in the tutorial and/or the source code comments.

gprs1809:
[...]

PR #309, opened 2022-01-11
Mention the remaining converters in the converters tutorial

garrison:
### Summary

I've added the remaining converters to the tutorial and standardized usage in a few places.  I was motivated to formulate a PR when I realized that `LinearInequalityToPenalty` is not the same thing as `LinearEqualityToPenalty` (it was added later, in #157).

### Details and comments

Per https://github.com/Qiskit/qiskit-optimization/issues/84#issuecomment-1008211844, I'm not sure if it matters the order in which the converters that `QuadraticProgramToQubo` wraps are listed.

### DIFF
--- a/docs/tutorials/02_converters_for_quadratic_programs.ipynb
+++ b/docs/tutorials/02_converters_for_quadratic_programs.ipynb
@@ -17,10 +17,13 @@
     "\n",
     "To map a problem to the correct input format, the optimization module of Qiskit offers a variety of converters. In this tutorial we're providing an overview on this functionality. Currently, Qiskit contains the following converters.\n",
     "\n",
-    "- `InequalityToEquality`: converts inequality constraints into equality constraints with additional slack variables.\n",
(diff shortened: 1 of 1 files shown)


woodsp-ibm:
I wonder whether listing the full set here is a good idea - as I guess happened now it gets out of data. Maybe it would be better to say it has converters and list some of them to show what they are and say for the full set of convertors refer to the API ref here https://qiskit.org/documentation/optimization/apidocs/qiskit_optimization.converters.html

garrison:
This PR started out as a single line where I added `LinearInequalityToPenalty`, which I think is worth mentioning alongside `LinearEqualityToPenalty`.  From there, I decided just to add everything that seemed to be missing, but I don't feel as strongly that all the other converters be listed.

t-imamichi:
This PR looks good because it just adds the missing converters.
But, as @woodsp-ibm pointed out, the API reference has the full list already, the list in this tutorial might be redundant.
If we introduce more converters in the future, it might be good to remove this list and revise this tutorial.

t-imamichi (approved):

PR #333, opened 2022-02-10
Fix up to_ising docstring

woodsp-ibm:
When referring someone to the `to_ising` method I noticed that the args were not documented and the return was not well formatted. 

image

- - - - - -

This PR adds the Args and fixes the return so it now looks like this 

image


t-imamichi (approved):
Thanks!

PR #361, opened 2022-04-12
Remove legacy providers interface

manoelmarques:
### Summary
See Qiskit/qiskit-terra#7886


woodsp-ibm:
We should probably have a release note for this.

t-imamichi (approved):

manoelmarques:
Terra has a release note but I will add one here too

manoelmarques:
Terra removed the legacy interfaces from its code and now the build fails. This PR needs to be merged.

PR #363, opened 2022-04-22
Update README.md

WhiteSymmetry:
### Summary

FutureWarning

### Details and comments

FutureWarning: adjacency_matrix will return a scipy.sparse array instead of a matrix in Networkx 3.0.  w = nx.adjacency_matrix(graph)
#pip install scipy -U (>=scipy-1.8.0)
#pip install networkx -U (>=networkx-2.8)

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,9 @@ elist = [(0, 1, 1.0), (0, 2, 1.0), (0, 3, 1.0), (1, 2, 1.0), (2, 3, 1.0)]
 graph.add_weighted_edges_from(elist)
 
 # Compute the weight matrix from the graph
-w = nx.adjacency_matrix(graph)
+#w = nx.adjacency_matrix(graph)
+w = nx.to_scipy_sparse_array(graph)
+#to_scipy_sparse_matrix(G, nodelist=None, dtype=None, weight='weight', format='csr')--> from_scipy_sparse_array(A, parallel_edges=False, create_using=None, edge_attribute='weight')
 
 # Formulate the problem as quadratic program
 problem = QuadraticProgram()


WhiteSymmetry:
> I see this is your first PR here. Its going to be helpful to outline/explain in the PR here your rationale for the change for the maintainers here. The summary and details were left blank. And eventually the change would just be the new code and any new comment/explanation if applicable, and not keep old code there etc commented out.

Thanks,
FutureWarning: adjacency_matrix will return a scipy.sparse array instead of a matrix in Networkx 3.0.  w=nx.adjacency_matrix(graph).

woodsp-ibm:
The sample in the readme is checked under CI to ensure it continues to work. With the above change it fails under Python 3.7
```

      File "/home/runner/work/qiskit-optimization/qiskit-optimization/test/test_readme_sample.py", line 62, in test_readme_sample
    self.fail(str(ex))

      File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/unittest/case.py", line 693, in fail
    raise self.failureException(msg)

    AssertionError: module networkx has no attribute to_scipy_sparse_array
```
as the latest networkx 2.8 requires Python 3.8 and above, so under 3.7 pip falls back to installing an older networkx where 3.7 was still supported.

As far as I can see the future warning just says that the return type is going to be switched from a matrix to an array to presumably notify anyone that may need to adapt their code. It seems that here its fine with the matrix, and also the array since you did that explicitly and the code still works ok. So when the adjacency_matrix code is changed from matrix to array in 3.0 the above should still work fine. I cannot see any other deprecation warning to suggest the adjacency_matrix  function is going away. Given the issue with 3.7 and the fact that it seems the code will continue to work I am wondering if there is really any need to do anything at present. @t-imamichi Any thoughts?

WhiteSymmetry:
> The sample in the readme is checked under CI to ensure it continues to work. With the above change it fails under Python 3.7
> 
> ```
> 
>       File "/home/runner/work/qiskit-optimization/qiskit-optimization/test/test_readme_sample.py", line 62, in test_readme_sample
>     self.fail(str(ex))
> 
>       File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/unittest/case.py", line 693, in fail
>     raise self.failureException(msg)
> 
>     AssertionError: module networkx has no attribute to_scipy_sparse_array
> ```
> 
> as the latest networkx 2.8 requires Python 3.8 and above, so under 3.7 pip falls back to installing an older networkx (2.6.3) where 3.7 was still supported and where that function does not exist.
> 
> As far as I can see the future warning just says that the return type is going to be switched from a matrix to an array to presumably notify anyone that may need to adapt their code. It seems that here its fine with the matrix, and also the array since you did that explicitly and the code still works ok. So when the adjacency_matrix code is changed from matrix to array in 3.0 the above should still work fine. I cannot see any other deprecation warning to suggest the adjacency_matrix function is going away. Given the issue with 3.7 and the fact that it seems the code will continue to work I am wondering if there is really any need to do anything at present. @t-imamichi Any thoughts?

Although the old code remains in place, for those who make updates every day, you can add the new code as an explanation underneath so that they are saved in the search for code for the update.

woodsp-ibm:
> Although the old code remains in place, for those who make updates every day, you can add the new code as an explanation underneath so that they are saved in the search for code for the update.

I think all the future warning says is that the return type will change. The new code will be what the adjacency_matix call will become internally, as far as I can tell, from the networkx source code. So the code as it was works now, and will work in the future. Its just that at present it also has the warning about the upcoming change - I expect in 3.0 the warning will be removed and the function internally will call the new ...sparse_array method instead of ...sparse_matrix. There is a TODO in the networkx code around that change. So in practice this change will make no difference and this sample will continue to work, just as it works now, as you show. Perhaps in a code sample,  having the future warning being emitted, may cause concern, as I guess it did for you. A comment could be added to note this perhaps. But since Qiskit supports 3.7 still whatever is done this sample has to work there as well. Lets see what @t-imamichi thoughts are too.

WhiteSymmetry:
> > Although the old code remains in place, for those who make updates every day, you can add the new code as an explanation underneath so that they are saved in the search for code for the update.
> 
> I think all the future warning says is that the return type will change. The new code will be what the adjacency_matix call will become internally, as far as I can tell, from the networkx source code. So the code as it was works now, and will work in the future. Its just that at present it also has the warning about the upcoming change - I expect in 3.0 the warning will be removed and the function internally will call the new ...sparse_array method instead of ...sparse_matrix. There is a TODO in the networkx code around that change. So in practice this change will make no difference and this sample will continue to work, just as it works now, as you show. Perhaps in a code sample, having the future warning being emitted, may cause concern, as I guess it did for you. A comment could be added to note this perhaps. But since Qiskit supports 3.7 still whatever is done this sample has to work there as well. Lets see what @t-imamichi thoughts are too.

In a few months, everyone will inevitably have to change. Until then, you can postpone it. You'll change it when the time comes.

t-imamichi:
Thank you for the heads up about FutureWarning of networkx. The example uses networkx to just generate an adjacency matrix.
It's easy to replace that part without using networkx. 
I made a PR to do that. https://github.com/Qiskit/qiskit-optimization/pull/365

t-imamichi:
@WhiteSymmetry What do you think of #365?

woodsp-ibm:
> In a few months, everyone will inevitably have to change. Until then, you can postpone it. You'll change it when the time comes.

Just to comment to this. As far as I can see no change is needed at all. The code works now and will work in the future when adjacency_matrix return is changed (you proved that by doing explicitly what I believe what adjacency_matrix will be become in 3.0 when its return type is changed). The fact is that the code does not depend on matrix, or sparse array as the return type and works with either as you showed. Networkx is issuing a warning now in case you do.

Anyway @t-imamichi provided an alternative that avoids using a graph package altogether which maybe nicer as it uses a docplex model.

t-imamichi:
@WhiteSymmetry Do you have any comment on #365?  If not, I will merge #365 and close this PR.

WhiteSymmetry:
> @WhiteSymmetry Do you have any comment on #365? If not, I will merge #365 and close this PR.

no comment

PR #365, opened 2022-04-25
Refactor the example in readme to remove networkx

t-imamichi:
### Summary
Refactor the example in readme to remove networkx.

Fixes #363


woodsp-ibm (approved):
I guess this is a nicer example in creating a model rather than starting out from a graph package.

PR #369, opened 2022-05-03
Adds informative text formats to `repr` and `str` for problem related objects (part 2/3 refactor text representation of objects)

t-imamichi:
### Summary

This adds informative text formats to `repr` and `str` for the following classes.
- LinearConstraint
- LinearExpression
- QuadraticConstraint
- QuadraticExpression
- QuadraticObjective
- Variable

### Details and comments

`repr` and `str` of `QuadraticProgram` and `OptimizationResult` will be updated in the next PR because they are used in tutorials and I have to update all tutorials. This PR does not affect existing tutorials.

The text of the formats is one line to be IDE friendly.

Vscode
image

Pycharm
image

I added `__repr__` and `__str__` to comply with Python's convention.

> `object.__repr__(self)`
> Called by the repr() built-in function to compute the “official” string representation of an object. If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.
https://docs.python.org/3/reference/datamodel.html#object.__repr__

> `object.__str__(self)`
> Called by str(object) and the built-in functions format() and print() to compute the “informal” or nicely printable string representation of an object. The return value must be a string object.
https://docs.python.org/3/reference/datamodel.html#object.__str__

### DIFF
--- a/qiskit_optimization/problems/linear_constraint.py
+++ b/qiskit_optimization/problems/linear_constraint.py
@@ -82,14 +82,14 @@ class LinearConstraint(Constraint):
 
     def __repr__(self):
         # pylint: disable=cyclic-import
-        from ..translators.prettyprint import _expr2str, DEFAULT_TRUNCATE
+        from ..translators.prettyprint import expr2str, DEFAULT_TRUNCATE
 
-        lhs = _expr2str(linear=self.linear, truncate=DEFAULT_TRUNCATE)
-        return f"<LinearConstraint: {lhs} {self.sense.label} {self.rhs} '{self.name}'>"
+        lhs = expr2str(linear=self.linear, truncate=DEFAULT_TRUNCATE)
+        return f"<{self.__class__.__name__}: {lhs} {self.sense.label} {self.rhs} '{self.name}'>"
 
     def __str__(self):
         # pylint: disable=cyclic-import
-        from ..translators.prettyprint import _expr2str
+        from ..translators.prettyprint import expr2str
 
-        lhs = _expr2str(linear=self.linear)
+        lhs = expr2str(linear=self.linear)
         return f"{lhs} {self.sense.label} {self.rhs} '{self.name}'"
--- a/qiskit_optimization/problems/linear_constraint.py
+++ b/qiskit_optimization/problems/linear_constraint.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2019, 2021.
+# (C) Copyright IBM 2019, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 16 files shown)


qiskit_optimization/problems/variable.py
```diff
@@ -148,3 +148,13 @@ def as_tuple(self) -> Tuple[str, Union[float, int], Union[float, int], VarType]:
...
+
+    def __repr__(self):
+        return f"<Variable: {str(self)}>"
+
+    def __str__(self):
+        if self._vartype == VarType.BINARY:
+            return f"{self._name} ({self._vartype.name.lower()})"
+        lowerbound = "" if self._lowerbound == -INFINITY else f"{self._lowerbound} <= "
+        upperbound = "" if self._upperbound == INFINITY else f" <= {self._upperbound}"
+        return f"{lowerbound}{self._name}{upperbound} ({self._vartype.name.lower()})"
```
woodsp-ibm, line 160: Just a comment. the `as_tuple` method above acts on similar values but does it via the getters  rather than direct access to the instance vars. Was there a specific reason to do this and not use the getters here i.e `self._name` versus `self.name`. I must admit when I have getters I tend to use these and not go direct to the instance vars that may be backing it. Since this is more an internal representation maybe that was the thought in being more direct.

releasenotes/notes/add-repr-str-9a33af0a6ffdca6e.yaml
```diff
@@ -0,0 +1,13 @@
+---
+upgrade:
```
woodsp-ibm, line 2: Is this an upgrade? or a feature. I thought upgrade were things an end user needed to do to upgrade their installation environment in some way to be compatible with the new release.

t-imamichi: I see. It then should be `features`. I will update it.

qiskit_optimization/problems/variable.py
t-imamichi: I don't have any specific reason to use `self._name` instead of `self.name`. I used `self._name` because `__str__` and `__repr__` are methods and are allowed to access the instance variable directly. I don't have a strong opinion on the choice of `self._name` or `self.name`. I follow your recommendation.

qiskit_optimization/problems/linear_constraint.py
```diff
@@ -79,3 +79,17 @@ def evaluate(self, x: Union[ndarray, List, Dict[Union[int, str], float]]) -> flo
             The left-hand-side of the constraint given the variable values.
         """
         return self.linear.evaluate(x)
+
+    def __repr__(self):
+        # pylint: disable=cyclic-import
+        from ..translators.prettyprint import _expr2str, DEFAULT_TRUNCATE
+
+        lhs = _expr2str(linear=self.linear, truncate=DEFAULT_TRUNCATE)
+        return f"<LinearConstraint: {lhs} {self.sense.label} {self.rhs} '{self.name}'>"
```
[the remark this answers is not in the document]
t-imamichi: Thanks. It's a smart solution.

qiskit_optimization/problems/linear_constraint.py
```diff
@@ -79,3 +79,17 @@ def evaluate(self, x: Union[ndarray, List, Dict[Union[int, str], float]]) -> flo
             The left-hand-side of the constraint given the variable values.
         """
         return self.linear.evaluate(x)
+
+    def __repr__(self):
+        # pylint: disable=cyclic-import
+        from ..translators.prettyprint import _expr2str, DEFAULT_TRUNCATE
```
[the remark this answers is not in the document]
t-imamichi: Alright. I rename `_expr2str` into `expr2str`

It's hard to avoid cyclic import between prettyprint and problem objects because prettyprint needs to access these problem objects.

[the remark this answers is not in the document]
woodsp-ibm: Ok, I figured it was mots likely a non-trivial task to deal with the cyclic import.

qiskit_optimization/problems/variable.py
woodsp-ibm: I don't have a strong opinion either which is why I started things above with "Just a comment..."

At present the getter simply returns the instance var value unaltered so using the instance var is the exact same; and I guess that's unlikely to change. Ok, lets stay with what you have then.

woodsp-ibm (approved):

PR #371, opened 2022-05-09
Updates the link to the release planning

t-imamichi:
### Summary

* Updates the link to the release planning
* removes tailing white spaces
* updates a URL to be included in <...>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,7 @@
 First please read the overall project contributing guidelines. These are
 included in the Qiskit documentation here:
 
-https://qiskit.org/documentation/contributing_to_qiskit.html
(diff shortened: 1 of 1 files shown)


HuangJunye:
Thanks for adding the link to the release planning wiki!

woodsp-ibm (approved):
Links do not need to be in angle brackets <> in the markdown here, as you can see in the current contributing.md if you view it. Its fine with it too though - makes it different from all the rest of the repos though in that regard!

t-imamichi:
Yes, Github recognizes the link as Github's [autolinks extension](https://github.github.com/gfm/#autolinks-extension-). I think the usual [autolink](https://github.github.com/gfm/#autolinks-extension-) format using <...> is better for other tools. I merge it, but I'm fine to revert it if there is any issue.

PR #378, opened 2022-06-12
added some code and explanation for how solutions are interpreted using the interpret method, in the tutorial of converters

gprs1809:


t-imamichi:
Thanks. I take a look at it.

gprs1809:
Hi @t-imamichi and @t-imamichi , thanks a lot! I am just a little confused. Are my additions visible?

t-imamichi:
Sorry for taking so long to review this PR. I will take over it and update your PR based on the current code.

t-imamichi:
I updated this PR as follows.

1. replaced the cells that manually interpret solutions with simply applies `interpret` because we don't have to show the internal code of `interpret`.
2. modified the coeffs of the objective function so that the optimal solution is unique so that the result of the original problem and that of the converted problem are the same.

woodsp-ibm:
As this is being updated, I see that this talks about VQE and QAOA in a couple of places, one at the start and one near to the end. Do we want to change this to `SamplingVQE` ? I know VQE is more known and maybe you want to leave it more generic like that - but of course the qiskit algorithms VQE can no longer used as since the change to primitives.

t-imamichi:
We refer to VQE as a general algorithm name. So, I think it's fine to leave it. We describe `SamplingVQE` in 03_minimum_eigen_optimizer.ipynb.

woodsp-ibm:
> We refer to VQE as a general algorithm name. So, I think it's fine to leave it.

Ok, since it docs, if it we do get any feedback its easy enough to adjust at any time.

PR #406, opened 2022-08-11
Enable inter-documentation links to be clickable

manoelmarques:
### Summary
Similar to Qiskit/qiskit-nature#783


woodsp-ibm:
How about adding networkx to the list `https://networkx.org/documentation/stable` that looks like it should work. And docplex too `https://ibmdecisionoptimization.github.io/docplex-doc` likewise.

woodsp-ibm (approved):

PR #427, opened 2022-10-26
Fix #426

### LINKED ISSUE #426: `tsp.interpret` returns nested lists
The return type of `tsp.interpret` seems quite strange: https://github.com/Qiskit/qiskit-optimization/blob/3276eaa9917e03923a76390050ed5cea2ecae6e6/qiskit_optimization/applications/tsp.py#L65

Sometimes when running tsp.interpret on a valid result, one can get complete nonsense like `[4, [2, 3], [], 1]`. In a project of mine, I solved this with `np.hstack` (see [my code here](https://github.com/rpitasky/qiskit-gcode-path-optimization/blob/d086257f18ef63509e64099ddfa59a0fe13d7eec/main.py#L150)) but it's strange to me that this is a problem at all.

I would expect a simple one-dimensional array of indices, and the tutorial does too! See https://github.com/Qiskit/qiskit-optimization/blob/3276eaa9917e03923a76390050ed5cea2ecae6e6/docs/tutorials/06_examples_max_cut_and_tsp.ipynb?short_path=a29aa13#L852.

After a little bit of poking around in the code, it seems to me like the iteration in the `tsp.interpret` function is handled wrong.

I would not mind fixing this if I get confirmation that this is an actual issue and not me misinterpreting something :)

rpitasky:
### Summary

Fixes #426 by removing a condition that confused the `extend` and `append` list methods. This should not affect any tests that wouldn't have failed already.

### Details and comments

See #426 for more details.

### DIFF
--- a/qiskit_optimization/applications/tsp.py
+++ b/qiskit_optimization/applications/tsp.py
@@ -62,7 +62,7 @@ class Tsp(GraphOptimizationApplication):
 
     def interpret(
         self, result: Union[OptimizationResult, np.ndarray]
-    ) -> List[Union[int, List[int]]]:
+    ) -> List[int]:
         """Interpret a result as a list of node indices
 
         Args:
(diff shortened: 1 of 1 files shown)


rpitasky:
@woodsp-ibm I fixed the style error, could you rerun the workflow?

rpitasky:
This problem is deeper than I thought, closing this PR until I can figure out what the problem is.

Issue #425, opened 2022-10-13
Support new primitive based algorithms.


woodsp-ibm:
### What is the expected enhancement?

Support primitive based algorithms such as the new QAOA and SamplingVQE etc accordingly.

Update GroverOptimizer to move away from QuantumInstance use, potentially by more directly using Grover from Terra.

Update tutorials correspondingly

Update readme sample

Migration guide as applicable

- [x] MinimumEigenOptimizer (@t-imamichi) #436
- [x] GroverOptimizer (@a-matsuo) #437
- [x] documentation #436
- [x] migration guide / tutorials #448

t-imamichi:
Tutorials and migration guide have just been merged #448. So, I think we can close this issue.

woodsp-ibm:
> So, I think we can close this issue.

Yes. I think so.

Maybe though, if you note above I mentioned for the GroverOptimzer the direct use of Grover algorithm from Terra. When the GroverOptimzer was originally done the Grover at that time was not as flexible. But now its likely that the Grover algorithm can be used here rather than the custom logic/circuits here. There was no time to look into this though for the current release but it may be worth raising a separate issue to cover this so it can be looked at later.

t-imamichi:
Thank you for reminding me of `GroverOptimizer`. Is this https://github.com/Qiskit/qiskit-optimization/pull/437/files#r1024573314 the discussion?

woodsp-ibm:
That certainly was the most recent discussion, yes. The idea has been floating around for a while though.

PR #436, opened 2022-11-11
Update optimizers to allow new primitives-based algorithms

t-imamichi:
### Summary

Add support of the new primitive-based algorithms to optimizers
- [x] `MinimumEigenOptimizer`
- [x] `GroverOptimizer`
- [x] `WarmStartQAOAOptimizer`

addresses #425

- [x] update codes
- [x] update tests
- [x] update docs
- [x] add reno
  - [x] add prelude 

Note that the update of tutorials will be a separate PR to minimize the size of PR.

### Details and comments

I copied some tests with legacy solvers to `test/algorithms/legacy` dir so that we can easiliy remove them when the legacy algorithms are deprecated.


qiskit_optimization/algorithms/minimum_eigen_optimizer.py
```diff
@@ -137,11 +151,18 @@ def __init__(
             TypeError: When one of converters has an invalid type.
             QiskitOptimizationError: When the minimum eigensolver does not return an eigenstate.
         """
-
+        if isinstance(min_eigen_solver, LegacyMinimumEigensolver):
```
woodsp-ibm, line 154: I am not sure whether we need a warning here. The legacy algos in Terra are pending deprecated and will be deprecated/removed, when creating one of these a user should already see a warning about them going away. We can have a warning here too, its just that the user should have already seen they should be using the newer ones.

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
```diff
@@ -137,11 +151,18 @@ def __init__(
...
+                f"min_eigen_solver as {LegacyMinimumEigensolver.__name__}",
+                new_name=f"min_eigen_solver as {SamplingMinimumEigensolver.__name__} "
+                f"or {NumPyMinimumEigensolver.__name__}",
+            )
         if not min_eigen_solver.supports_aux_operators():
             raise QiskitOptimizationError(
                 "Given MinimumEigensolver does not return the eigenstate "
-                + "and is not supported by the MinimumEigenOptimizer."
+                "and is not supported by the MinimumEigenOptimizer."
             )
```
woodsp-ibm, line 166: The supports_aux_operators used to imply that the algorithm had the state to measure aux ops. With the estimator this is no longer the case such that while new VQE can support_aux_operators it no longer has a eigenstate field. Nowdays if you want the state you would have to use the ansatz and optimal parameters from the result to compute it yourself, if that applies, say with a Sampler.

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
```diff
@@ -11,24 +11,38 @@
...
 )
-from ..exceptions import QiskitOptimizationError
-from ..converters.quadratic_program_to_qubo import (
-    QuadraticProgramToQubo,
-    QuadraticProgramConverter,
-)
-from ..problems.quadratic_program import QuadraticProgram, Variable
+
+MinimumEigensolver = Union[
+    SamplingMinimumEigensolver, NumPyMinimumEigensolver, LegacyMinimumEigensolver
```
woodsp-ibm, line 41: So while the old legacy VQE is supported the new one is not. They have to be SamplingMinimumEigensolvers (QAOA, SamplingVQE) - the only exception is the classical NumPyMinimumEigensolver for which we do not have a version conforming to that sampling interface. Maybe we should have some specific note in the release note and/or migration guide if the latter is being done, around VQE as it was used before MinEigOptimizer tutorials.

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
t-imamichi: Thank you for your suggestion. I will remove the warning message.

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
t-imamichi: Yes, it is confusing that users need to use `SamplingVQE` instead of `VQE` of the new algorithms. I will mention it in reno.

t-imamichi: I mentioned it in reno.

qiskit_optimization/deprecation.py
```diff
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2021, 2022.
```
t-imamichi, line 3: I modified this flle in the PR but I reverted the change after all. The copyright.py still updates the year. Should I follow the tool, @manoelmarques?

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
t-imamichi: Yes, `MinimumEigenOptimizer` requires sampler-based solver `SamplingVQE` instead of `VQE` due to this reason.

qiskit_optimization/deprecation.py
manoelmarques: If the file is being commited, then you should follow the tool unless you reverted the commit and removed the file from that commit and committed again.

t-imamichi: Thank you for your information. I could not revert the commit because it includes changes of other files too. So, I follow the tool to update year.

releasenotes/notes/add-primitives-support-31af39549b5e66e3.yaml
```diff
@@ -0,0 +1,24 @@
+---
```
woodsp-ibm, line 1: Could we do some prelude section here that has a high level summary - i.e. more or less that Qiskit Optimization has been updated to support the primitive based algorithms introduced in Qiskit Terra 0.22. The older algorithms that took or used quantum instance, are still supported here until they are removed from Qiskit Terra in a future release. We encourage you to migrate to using the new primitive based algorithms.

Or something like that which just gives a high level statement. Will we do migration tutorial that shows examples of how it would have been done and how its done now with primitive ones. Since the optimization code is the same I think its really down to the algorithm creation I believe. It maybe helpful to show this anyway in a couple of examples. Maybe that could be added in the PR with the tutorial updates if it seems useful?

releasenotes/notes/add-primitives-support-31af39549b5e66e3.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - |
+    The :class:`~.GroverOptimizer` class has a new keyword argument, ``sampler`` which is
+    used to run the algorithm using an instance of the :class:`~.BaseSampler`
+    interface to calculate the results. This new argument supersedes
+    the ``quantum_instance`` argument and accordingly, ``quantum_instance``
+    is pending deprecation and will be deprecated and subsequently removed in
+    future releases.
+  - |
+    The :class:`~.MinimumEigenOptimizer` class takes the primitives-based algorithms
+    (``qiskit.algorithms.minimum_eigensolvers.SamplingVQE`` and
```
woodsp-ibm, line 12: In theory it supports any subclass of `SamplingMinimumEigensolver` right (that says it supports aux_ops ie has a valid eigenstate value in the result. `SamplingVQE` is all there is for now but it could be cited as an example of the former.

t-imamichi: You are right. It's my mistake. I will fit it.

t-imamichi:
Thank you for your review.

I will update the terra version to `qiskit-terra>=0.22.0`, catch `PendingDuplicationWarning` in the legacy tests, and add prelude in reno.

I didn't notice the adjustment of our codes for matplotlib. Could you make an issue of adjustments we need?

woodsp-ibm:
> I will update the terra version to `qiskit-terra>=0.22.0`, catch `PendingDuplicationWarning` in the legacy tests, and add prelude in reno.

In Nature I believe we used >= 0.22.* so as it will work with release candidates too, as Terra is doing that now.

When changing the code you might like to catch both PendingDeprecation and Deprecation  to save changing the code as Terra should switch the legacy algos over to that for its next release in January at some point - would save having to go back in an edit them again. Since you have the legacy tests there it should be easy to do it in setup and teardown.

> I didn't notice the adjustment of our codes for matplotlib. Could you make an issue of adjustments we need?

If you look at the build details, at the `Deprecation_Messages_and_Coverage` the `Combined Deprecation Messages` part of the CI output shows all the messages that were recorded. I did not look at them too hard, just noticed the `matplotlib` one, and also a `networkx` one I think. The matplotlib one (Pending deprecation so its not yet deprecated) is

> applications\bin_packing.py:121: PendingDeprecationWarning: The get_cmap function will be deprecated in a future version. Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap(obj)`` instead.

t-imamichi:
Thanks. I use requirements.txt to use `qiskit-terra>=0.22.*`.

I updated tests to catch pending deprecation warnings except VQEClient and QAOAClient. I think I need to sync them with qiskit-nature's VQEClient, which still use the conventional classes [link](https://github.com/Qiskit/qiskit-nature/blob/b4a330b2b479602ab45526db94d023851f5fb14d/qiskit_nature/runtime/vqe_client.py#L24).

I also updated networkx and matploitlib (pending) deprecation warnings because they are just one line change.

t-imamichi:
Yes, I plan to remove VQEProgram and QAOAProgram in a separate PR.

> Also for the runtime, the former versions of these VQEProgram and QAOAProgram were deprecated in 0.3.0 so can be removed in this 0.5.0 release.

I also notice that networkx 2.2 and 2.3 does not support Python 3.8 or later. I need to update requirements.txt. So, I will work on the fix in a separeta PR.

> I had wondered about this and what versions of the libraries might be needed for the replacement method(s) and whether requirements.txt needed updating. But it looks like us supporting 3.7 may well need to have an older matplotlib where the new method does not yet exist. Maybe we live with the pending deprecation for now, or do some more dynamic check/code for the new method and fall back to the old way if needed.

releasenotes/notes/add-primitives-support-31af39549b5e66e3.yaml
```diff
@@ -0,0 +1,31 @@
+---
+prelude: >
+    Qiskit Optimization 0.5 supports the new algorithms introduced in Qiskit Terra 0.22
+    which in turn rely on the `Qiskit Primitives <https://qiskit.org/documentation/apidoc/primitives.html>`_.
+    Qiskit Optimization 0.5 still supports the conventional algorithms based on ``QuantumInstance``.
+    But they will be deprecated in the future release.
```
woodsp-ibm, line 6:
```suggestion
    Qiskit Optimization 0.5 still supports the former algorithms based on ``QuantumInstance``,
    but they will be deprecated and then removed, along with the support here, in future releases.
```
I switched `conventional` to `former` to more indicate it was the older pre-existing algos. And the last bit I changed a little to add that these will be removed too along with the support code here in optimization.

releasenotes/notes/add-primitives-support-31af39549b5e66e3.yaml
woodsp-ibm: I made a small suggestion but I think its good and captures what is new in the release as a high level overview for the reader.

releasenotes/notes/add-primitives-support-31af39549b5e66e3.yaml
t-imamichi: Thanks. It's more sophisticated.

t-imamichi:
The unit tests fail with the following message. But I'm not sure how to fix it. Do you have any idea, @manoelmarques?

```
/home/runner/work/qiskit-optimization/qiskit-optimization/docs/release_notes.rst:36: WARNING: Inline literal start-string without end-string.
```

releasenotes/notes/add-primitives-support-31af39549b5e66e3.yaml
```diff
@@ -0,0 +1,24 @@
...
+    The :class:`~.MinimumEigenOptimizer` class takes the primitives-based algorithms
+    (``qiskit.algorithms.minimum_eigensolvers.SamplingMinimumEigensolver`` and
+    ``qiskit.algorithms.minimum_eigensolvers.NumPyMinimumEigensolver``)
+    as ``min_eigen_solver`` argument.
+    The former algorithm ``qiskit.algorithms.minimum_eigen_solvers.MinimumEigensolver``
+    is pending deprecation and will be deprecated and subsequently removed in future releases.
+    Note that ``qiskit.algorithms.minimum_eigensolvers.SamplingVQE`` supersedes
+    ``qiskit.algorithms.VQE`` for :class:`~.MinimumEigenOptimizer`.
+    ``qiskit.algorithms.minimum_eigensolvers.NumPyMinimumEigensolver`` also supersedes
+    ``qiskit.algorithms.minimum_eigen_solvers.NumPyMinimumEigensolver`.
```
woodsp-ibm, line 19:
```suggestion
    ``qiskit.algorithms.minimum_eigen_solvers.NumPyMinimumEigensolver``.
```
Perhaps the release note is failing due to the mistmatch here - starts with \`\` but ended with a single \`

t-imamichi: Thanks. You are right.

t-imamichi:
While I update the tutorial #442, I notice that `OptimizationApplication.sample_most_likely` needs to be updated. Since the method does not have a test, I didn't notice it. I will add a unit test of the method too.

t-imamichi:
I have updated `OptimizationApplication.sample_most_likely`. It gets ready for review again.

releasenotes/notes/add-primitives-support-31af39549b5e66e3.yaml
```diff
@@ -0,0 +1,24 @@
...
+prelude: >
+    Qiskit Optimization 0.5 supports the new algorithms introduced in Qiskit Terra 0.22
+    which in turn rely on the `Qiskit Primitives <https://qiskit.org/documentation/apidoc/primitives.html>`_.
+    Qiskit Optimization 0.5 still supports the former algorithms based on ``QuantumInstance``,
+    but they will be deprecated and then removed, along with the support here, in future releases.
+
+features:
+  - |
+    The :class:`~.MinimumEigenOptimizer` class takes the primitives-based algorithms
+    (``qiskit.algorithms.minimum_eigensolvers.SamplingMinimumEigensolver`` and
```
woodsp-ibm, line 11: Can we change all the highlighted qiskit.algorithm entries here over  to \:class\: references. These should resolve through intersphinx right so that they will be links not just highlighted entries.

releasenotes/notes/grover-opt-primitive-de82d051d6cee2e4.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    The :class:`~.GroverOptimizer` class has a new keyword argument, ``sampler`` which is
+    used to run the algorithm using an instance of the :class:`~.BaseSampler`
```
woodsp-ibm, line 5: > :class:`~.BaseSampler`

This needs to be a fully qualified path over to the BaseSampler in qiskit terra, otherwise its just highlighted in the output as it cannot be resolved like it is, since it's not part of optimization.

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -199,7 +237,7 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
         # Initialize oracle helper object.
         qr_key_value = QuantumRegister(self._num_key_qubits + self._num_value_qubits)
         orig_constant = problem_.objective.constant
-        measurement = not self.quantum_instance.is_statevector
+        measurement = not (self._quantum_instance and self._quantum_instance.is_statevector)
```
woodsp-ibm, line 240: `self._quantum_instance` this should ideally be an explicit test against None since that's the intent here I believe.

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -246,15 +284,19 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
                     threshold = optimum_value
 
                     # trace out work qubits and store samples
-                    if self._quantum_instance.is_statevector:
-                        indices = list(range(n_key, len(outcome)))
-                        rho = partial_trace(self._circuit_results, indices)
-                        self._circuit_results = cast(Dict, np.diag(rho.data) ** 0.5)
-                    else:
+                    if self._sampler:
```
woodsp-ibm, line 287: This should be an explicit test against None I think too.

qiskit_optimization/algorithms/optimization_algorithm.py
```diff
@@ -566,7 +568,25 @@ def generate_solution(bitstr, qubo, probability):
             )
 
         solutions = []
-        if isinstance(eigenvector, dict):
+        if isinstance(eigenvector, QuasiDistribution):
```
woodsp-ibm, line 571: I guess here it doesn't matter that we might have negative values, which I think might happen for an error mitigated result. Such values will be discarded anyway by the threshold. Maybe we don't need to bother about the converting to a nearest distribution - which though might change the result from here I guess.

releasenotes/notes/add-primitives-support-31af39549b5e66e3.yaml
t-imamichi: Sure. I will do it.

releasenotes/notes/grover-opt-primitive-de82d051d6cee2e4.yaml
t-imamichi: Thanks. I update it.

qiskit_optimization/algorithms/grover_optimizer.py
t-imamichi: @a-matsuo What do you think of this?

qiskit_optimization/applications/optimization_application.py
```diff
@@ -59,29 +61,54 @@ def _result_to_x(self, result: Union[OptimizationResult, np.ndarray]) -> np.ndar
...
         Returns:
             binary string as numpy.ndarray of ints.
+
+        Raises:
+            ValueError: if state_vector is not QuasiDistribution, Statevector,
+                np.ndarray, or dict.
         """
-        if isinstance(state_vector, (OrderedDict, dict)):
+        if isinstance(state_vector, QuasiDistribution):
+            probabilities = state_vector.binary_probabilities()
```
t-imamichi, line 80: Same discussion. https://github.com/Qiskit/qiskit-optimization/pull/436#discussion_r1033283358

qiskit_optimization/algorithms/optimization_algorithm.py
t-imamichi: `OptimizationAlgorithm` does not have any option to control `_eigenvector_to_solutions` method. So, we need to introduce an option, e.g., `nearest_probability`, to the constructor of `OptimizationAlgorithm`. What do you think?

qiskit_optimization/algorithms/grover_optimizer.py
a-matsuo: In the beginning of `solve()`, we have the following code to ensure only one of a sampler or quantum instance is given. Do you mean that still it's better to explicitly write like `self._sampler and not self._quanutm_instance`?
```
      if self._sampler is None and self._quantum_instance is None:
          raise ValueError("A quantum instance or sampler must be provided.")

      if self._quantum_instance is not None and self._sampler is not None:
          raise ValueError("Only one of quantum_instance or sampler can be passed, not both!")
```

woodsp-ibm: This is just about the part `not(self._quantum_instance..) above and below (if self._sampler) - the recommendation is when comparing against None to do an explicit test https://peps.python.org/pep-0008/#programming-recommendations. Its being done in other tests as I would expect - just not these places I noted. I don't mean change the logic around as such just make the test itself, which is doing it based on None or None - explicit rather than the bool conversion.

qiskit_optimization/algorithms/optimization_algorithm.py
t-imamichi: This PR is already a bit too big. So, I want to leave this option for another PR. Is it OK?

qiskit_optimization/algorithms/grover_optimizer.py
t-imamichi: Do you mean the following?
```suggestion
        measurement = self._quantum_instance is None or not self._quantum_instance.is_statevector
```

qiskit_optimization/algorithms/grover_optimizer.py
t-imamichi: Do you mean the following?
```suggestion
                    if self._sampler is not None:
```

releasenotes/notes/add-primitives-support-31af39549b5e66e3.yaml
t-imamichi: done

qiskit_optimization/algorithms/optimization_algorithm.py
woodsp-ibm: Sure, a separate PR to address this sounds fine. Gives a bit more time too to think/consider options.

qiskit_optimization/algorithms/grover_optimizer.py
woodsp-ibm: Yes that works

t-imamichi:
I applied Steve's suggestions. So, this PR is about to merge. The option of binary prob or nearest prob https://github.com/Qiskit/qiskit-optimization/pull/436#discussion_r1032564684 will be a separate PR.

t-imamichi:
I'm asking @garrison to take a look at this PR to avoid break of https://github.com/qiskit-community/prototype-qrao

t-imamichi:
According to Jim, this PR works fine with QRAO. So, I merge this PR.

PR #437, opened 2022-11-15
`GroverOptimizer` with `Sampler`

a-matsuo:
### Summary
Add support of `Sampler` to `GroverOptimizer`

Addresses #425

- [x] update codes
- [x]  update tests
- [x]  update docs
- [x]  add reno


t-imamichi:
Thank you, @a-matsuo. I'm thinking of integrating this PR to #436.

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -54,6 +57,7 @@ def __init__(
             Union[QuadraticProgramConverter, List[QuadraticProgramConverter]]
         ] = None,
         penalty: Optional[float] = None,
+        sampler: Optional[BaseSampler] = None,
```
woodsp-ibm, line 60: At one point I thought there had been discussion about changing things here so it used the Grover algorithm from Terra - I thought it was now possible since this was done long ago and the Grover implementation since then had become more capable where this could be done. Maybe as an alternate new path, while keeping the old more direct code this is more complex to achieve this time around - i.e. instead of a Sampler it would take a Grover (which itself would already be setup with a Sampler).

t-imamichi:
Since this PR is based on #436, I will merge this PR to #436.

qiskit_optimization/algorithms/grover_optimizer.py
a-matsuo: Thx for the reviewing. I agree with that it is nice to use the `Grover` class in Terra. I believe it's possible since the `Grover` class can receive functions for `iteration` and `is_good_state`. So, we can dynamically change the iteration of the Grover operator in the algorithm by squashing the current code into `iteration` and `is_good_state`. (This might be a good issue for the qiskit advocate mentorship program. I don't have time right now. )

For `GroverOptimizer`, we need to change the variables of `Grover` dynamically. So, we can pass something like `Grover(quantum_instance=quantum_instance)` or `Grover(sampler=sampler)`, but this doesn't make much difference, and I think passing a sampler looks better than passing `Grover(sampler=sampler)` to me.

t-imamichi:
I have merged it to #436. So I close this PR.

PR #447, opened 2022-11-30
[DNM] Add ScipyMilpOptimizer

Dran-Z:
### Summary

Add optimizer method for linear problem based on `scipy.optimize.milp` solver. This is a [QAMP-fall-2022](https://github.com/qiskit-advocate/qamp-fall-22/issues/2) project.

### DIFF
--- a/.pylintdict
+++ b/.pylintdict
@@ -111,6 +111,7 @@ maxcut
 maxfun
 maxiter
 mdl
+milp
 minimizer
 minimumeigenoptimizer
 multiset
--- a/qiskit_optimization/algorithms/__init__.py
+++ b/qiskit_optimization/algorithms/__init__.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2019, 2021.
+# (C) Copyright IBM 2019, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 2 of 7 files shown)


t-imamichi:
Thank you for your contribution.

I notice that scipy 1.8.0 dropped Python 3.7 support https://pypi.org/project/scipy/1.8.0/#files.
The latest scipy 1.9.3 available for Python 3.8-3.11 https://pypi.org/project/scipy/1.9.3/#files.
So, you need to extend `_optional` to have `_optional.HAS_SCIPY_MILP` and move `import scipy.optimize...` to the internal of methods `_generate_problem` and `solve` in a similar way  to gurobi_optimizer.py

Dran-Z:
Thanks for your comment. I will try to resolve the problem then.

t-imamichi:
Could you push the latest version?

Dran-Z:
> Could you push the latest version?

Sure! Done now

woodsp-ibm:
@Dran-Z The next version of Qiskit Optimization should release in a few weeks. Would you like to fix this up so its ready to be included in this release?

Since we have scipy in the requirement.txt `scipy>=1.4` having it as an optional install as well in setup.py seems weird. Python 3.7 support will end in a couple of months and then we can choose to list scipy>=1.9 as a requirement. Since the milp support is not going to work for Python 3.7 (no scipy version compatible) we should probably note this, and the min version needed for other Python versions i.e. 1.9. For the optional text we could just have the install as `pip install -U scipy` and drop the optional aspect from the setup.py.

t-imamichi:
If he cannot handle it, I will take over this work.

Dran-Z:
@t-imamichi Sorry I am recently quite occupied to prepare a manuscript. It should help a lot if you could take a look at it. Many thanks!

t-imamichi:
I will make another PR because @Dran-Z uses main branch of his repo for this PR.
If I update this main branch, it may cause some troubles.

t-imamichi:
I close this PR because #496 has been merged.

PR #450, opened 2022-12-02
Raise an error if `VQE` is given to `MinimumEigenOptimizer`

t-imamichi:
### Summary

This PR has `MinimumEigenOptimizer` raises an error if the new `VQE` is given because `MinimumEigenOptimizer` supports `SamplingVQE` instead.
I added a special error message to suggest `SamplingVQE` if `VQE` is given because it can often happen.

### DIFF
--- a/qiskit_optimization/algorithms/minimum_eigen_optimizer.py
+++ b/qiskit_optimization/algorithms/minimum_eigen_optimizer.py
@@ -154,8 +154,8 @@ class MinimumEigenOptimizer(OptimizationAlgorithm):
         """
         if isinstance(min_eigen_solver, VQE):
             raise TypeError(
-                "MinimumEigenOptimizer does not support VQE but SamplingVQE. "
-                "qiskit.algorithms.minimum_eigensolvers.SamplingVQE may help you."
+                "MinimumEigenOptimizer does not support this VQE. You can use  "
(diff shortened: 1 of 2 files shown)


qiskit_optimization/algorithms/minimum_eigen_optimizer.py
```diff
@@ -147,10 +148,26 @@ def __init__(
...
         Raises:
+            TypeError: If minimum eigensolver has an invalid type.
             TypeError: When one of converters has an invalid type.
             QiskitOptimizationError: When the minimum eigensolver does not return an eigenstate.
         """
-
+        if isinstance(min_eigen_solver, VQE):
+            raise TypeError(
+                "MinimumEigenOptimizer does not support VQE but SamplingVQE. "
+                "qiskit.algorithms.minimum_eigensolvers.SamplingVQE may help you."
```
woodsp-ibm, line 158:
```suggestion
                "MinimumEigenOptimizer does not support this VQE. You can use  "
                "qiskit.algorithms.minimum_eigensolvers.SamplingVQE instead."
```

Since it had SamplingVQE in short and in full following one from one another I altered to just leave the full one and changed text a little to suggest to make it more explicit they can use this instead.

woodsp-ibm (approved):

PR #465, opened 2023-01-23
Update the minimum version of Qiskit Terra

t-imamichi:
### Summary

Qiskit optimization requires Terra 0.22.4 that includes various fixes of sampling-based algorithms https://github.com/Qiskit/qiskit-terra/releases/tag/0.22.4 to run the tutorials properly.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-qiskit-terra>=0.22.*
+qiskit-terra>=0.22.4
 scipy>=1.4
 numpy>=1.17
 docplex>=2.21.207,!=2.24.231


manoelmarques:
Terra 0.23 is almost out. Why not require `>=0.23.*' ? This way it won't loose the ability to be installed with Terra RC's.

t-imamichi:
I made this PR because the updated tutorials require terra 0.22.4. 0.23.*  would work fine too. I don't have a strong opinion. `>=0.22.4` allows users to use 0.22 while `>=0.23.*` forces to upgrade to 0.23.

manoelmarques:
In the end RC releases are meant to be used to verify if Terra still has bugs before the final release. End users should be working with the final releases and not RC. If someone wants to use a Terra RC release, one can just install it. Having > 0.22.4 more accurately describes the minimum version required.

manoelmarques (approved):

t-imamichi:
Thank you for your information. I then merge this PR.

Issue #433, opened 2022-11-10
Replace opflow operators with quantum_info `SparsePauliOp`


t-imamichi:
### What is the expected enhancement?

As primitive-based algorithm is rolled out and opflow might be deprecated in the future, we might need to update Qiskit-optimization to replace opflow operators with `SparsePauliOp` of quantum_info.
For example, `to_ising` returns `PauliSumOp` and `from_ising` receives `PauliSumOp`.

woodsp-ibm:
I imagine we can transition over already to some degree if we choose.  For `from_ising` we could probably support conversion for both the opflow OperatorBase and the quantum_info BaseOperator with the former type being deprecated, so things can be transitioned over. For to_ising maybe add an additional parameter that presently defaults to producing an OperatorBase (PauliSumOp), so any users code still works, but set it via when optimization uses it internally so it goes over to producing an OperatorBase - SparsePauliOp. Maybe this new parameter will be removed when only OperatorBase is emitted. I guess as far as the optimization stack is concerned its use of to_ising is all internal and can just be switched based on the algorithm, whether it be older QI based ones, so users existing code continues to work or the new primitive based ones that can take a BaseOperator directly - at least it looks that way for the minimum eigen optimizer.  Of course the new primitive based algos still work with PaulisSumOp so switching at present is not strictly needed. But for the public API of to/from_ising it would probably be better to additionally do it a new way while supporting the old. Please note that with the new primitive Terra NumPy(Minimum)Eigensolver there is presently a bug Qiskit/qiskit-terra#9101

t-imamichi:
Thank you for the feedback. Yes, we have to update the code without breaking user codes. I intend to work on it when opflow is deprecated and the new primitive-based algorithms get stable.

PR #467, opened 2023-01-23
Update `to_ising` and `from_ising` to support `SparsePauliOp`

t-imamichi:
### Summary

- Adds `opflow` option to `to_ising` to control the output is `SparsePauliOp` or Opflow's operator.
- Allows `from_ising` to receive `quantum_info`'s operators.

This is part of #433.
The legacy minimum eigensolver of Terra supports only opflow (not `SparsePauliOp`).
We will deprecate the legacy minimum eigensover and opflow when they are deprecated in Terra.


qiskit_optimization/translators/ising.py
```diff
@@ -13,17 +13,20 @@
...
+from qiskit.quantum_info import Pauli, SparsePauliOp
+from qiskit.quantum_info.operators.base_operator import BaseOperator
+
 from qiskit_optimization.exceptions import QiskitOptimizationError
 from qiskit_optimization.problems.quadratic_program import QuadraticProgram
 
 
-def to_ising(quad_prog: QuadraticProgram) -> Tuple[OperatorBase, float]:
+def to_ising(
+    quad_prog: QuadraticProgram, quantum_info: bool = False
```
woodsp-ibm, line 28: I imagine this `quantum_info` parameter will get removed once only SparsePauliOp exists. Before then, at some point, we need to transition over to True by default where False is supported but deprecated. So for the release where we introduce this we have it still producing PauliSumOp by default, and somewhere I would think have a warning message saying the default will be changed to SparsePauliOp. When we do that passing False in, to produce a PauliSumOp would be deprecated. We probably should state this intent in the release note. Support both for now defaulting to PauliSumOp, the release after default to SparsePauliOp (deprecating PauliSumOp) and the release after removing that support leaving just SparsePauliOp supported with the flag here gone.

t-imamichi: Yes, you are right. I plan to remove opflow by 1) allow SparsePauliOp with this PR though the default is opflow, 2) change the default with SparsePauliOp, 3) remove opflow. It also makes sense that the warning message and release note mention the plan.

t-imamichi:
I updated the option name from `quantum_info` to `opflow` because I thought it is easier to understand why it will be deprecated when opflow is deprecated in terra.
I also changed the code to display a warning message if `opflow` option is not set explicitly.

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
```diff
@@ -222,7 +222,7 @@ def solve(self, problem: QuadraticProgram) -> MinimumEigenOptimizationResult:
         problem_ = self._convert(problem, self._converters)
 
         # construct operator and offset
-        operator, offset = problem_.to_ising()
+        operator, offset = problem_.to_ising(opflow=True)
```
woodsp-ibm, line 225: Where we use conversion methods internally, like here, and consume the operators ourselves, in this case via the min eigen solver, we could already move over to using SparsePauliOp now couldn't we. Is there any reason not to? Its what we would recommend right that people start the migration away from PauliSumOp (opflow) - ideally any new code would start out that way though of course its not the default yet for compatibility/migration reasons.

releasenotes/notes/support-sprase-pauli-op-00d29000d48c93ca.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    Updates :func:`~.to_ising` and :func:`~.from_ising` to support
+    ``qiskit.quantum_info.SparsePauliOp``.
+    Adds ``opflow`` option to :func:`~.to_ising` to control the output is
+    opflow's operator (if ``True``) or ``SparsePauliOp`` (if ``False``).
```
woodsp-ibm, line 7:
```suggestion
    opflow's operator (if ``True``) or ``SparsePauliOp`` (if ``False``).
```

qiskit_optimization/translators/ising.py
```diff
@@ -32,6 +36,8 @@ def to_ising(quad_prog: QuadraticProgram) -> Tuple[OperatorBase, float]:
 
     Args:
         quad_prog: The problem to be translated.
+        opflow: The output object is an OpFlow's operator if True.
```
woodsp-ibm, line 39: Juts checking on `is an OpFlow's operator`  - the output here is always a PauliSumOp right, may be we can say then `an opflow PauliSumOp`. In practice I think it has to be a PauliSumOp since thats the only opflow operator listed by primitives as supported i.e. `BaseOperator | PauliSumOp`

qiskit_optimization/translators/ising.py
```diff
@@ -42,6 +48,15 @@ def to_ising(quad_prog: QuadraticProgram) -> Tuple[OperatorBase, float]:
             in the problem.
         QiskitOptimizationError: If constraints exist in the problem.
     """
+    if opflow is None:
+        opflow = True
+        warn(
+            "`opflow` option of `to_ising` is not set explicitly. "
+            "It is set True now, but the default value will be changed to False. "
+            "We suggest setting `opflow=True` explicitly like `to_ising(qp, opflow=True)`.",
+            stacklevel=2,
```
woodsp-ibm, line 57: I'll comment here but it partly relates to the text in the release note. Here we default now to True so as to be backward compatible, but this default will be changed to False in a future (the next) release. The use of opflow operators will be deprecated at that point and will not be the default - presumably at that time, here we will detect anyone setting True explicitly and warn to change to False, or preferably remove their explicit setting since the opflow parameter will be removed after the deprecation period. We should recommend moving away from using opflow operators, which can be done right now by passing False here.

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
t-imamichi: Unfortunately, we cannot move to `SparsePauliOp` here because `MinimumEigenOptimizer` need to accept the legacy minimum eigensolvers that can deal with only opflow.

qiskit_optimization/translators/ising.py
```diff
@@ -148,6 +172,11 @@ def from_ising(
         QiskitOptimizationError: if any Pauli term has an imaginary coefficient
         NotImplementedError: If the input operator is a ListOp
     """
+    if isinstance(qubit_op, BaseOperator):
```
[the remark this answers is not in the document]
t-imamichi: Good idea. I will add a pending deprecation warning.

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
woodsp-ibm: Ah right, so it has to stay this way, even if the default of the to_ising gets changed, until the support for the legacy algorithms is removed.

qiskit_optimization/translators/ising.py
t-imamichi: I updated the code so that the return object can be always PauliSumOp with `opflow=True`.

qiskit_optimization/translators/ising.py
t-imamichi: Added a warning if `opflow` is not set explicitly

qiskit_optimization/translators/ising.py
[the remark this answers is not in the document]
t-imamichi: Added a pending deprecation warning.

a-matsuo (commented):
I reviewed the code changes thoroughly and everything looks good to me!

a-matsuo (approved):
I reviewed the code changes thoroughly and everything looks good to me!

qiskit_optimization/translators/ising.py
```diff
@@ -131,6 +150,12 @@ def from_ising(
     i-th variable is mapped to i-th qubit.
     See https://github.com/Qiskit/qiskit-terra/issues/1148 for details.
 
+    .. note::
+
+        The ``qubit_op`` argument can currently accept Opflow operators, but it has been
+        superseded by ``SparsePauliOp``.
+        Opflow will be deprecated in a future release and subsequently removed after that.
```
woodsp-ibm, line 157:
```suggestion
        The ``qubit_op`` argument can currently accept Opflow operators (OperatorBase type), but have been
        superseded by Qiskit Terra quantum info BaseOperators such as  ``SparsePauliOp``.
        Opflow operator support will be deprecated in a future release and subsequently removed after that.
```
Here is accepts all quantum info BaseOperator types right. Most often it will be SparsePauliOp bit its not restricted to this. I will also note that we can make these class references since we have the intersphinx with Terra.

qiskit_optimization/translators/ising.py
```diff
@@ -148,9 +173,24 @@ def from_ising(
...
+    if isinstance(qubit_op, BaseOperator):
+        if not isinstance(qubit_op, SparsePauliOp):
+            qubit_op = SparsePauliOp(qubit_op)
 
+    # opflow
+    if isinstance(qubit_op, OperatorBase):
+        warn(
+            "The `qubit_op` argument can currently accept Opflow operators, but it has been "
+            "superseded by `SparsePauliOp`. "
+            "Opflow will be deprecated in a future release and subsequently removed after that.",
```
woodsp-ibm, line 186: I did not alter the text here as per he above suggestion. But if the suggestion seems better then perhaps this should be the same.

releasenotes/notes/support-sprase-pauli-op-00d29000d48c93ca.yaml
```diff
@@ -0,0 +1,13 @@
+---
+features:
+  - |
+    Updated :func:`~.to_ising` to support ``qiskit.quantum_info.SparsePauliOp``.
+    Added ``opflow`` argument to :func:`~.to_ising` to control the output is
+    opflow operator (if ``True``) or ``SparsePauliOp`` (if ``False``).
+    The default value of ``opflow`` argument is currently ``True``, but it will
+    first be changed to ``False`` and then deprecated in future releases.
```
woodsp-ibm, line 8:
```suggestion
    Updated :func:`~.to_ising` to support optionally returning ``qiskit.quantum_info.SparsePauliOp``.
    This is achieved by a new ``opflow`` argument to :func:`~.to_ising` to control if the output is
    the opflow operator, as has been done in the past,  (if ``True``) or ``SparsePauliOp`` (if ``False``).
    The default value of ``opflow`` argument is currently ``True``, but it will
    first be changed to ``False``. The parameter will be deprecated and removed in a future release.
```

releasenotes/notes/support-sprase-pauli-op-00d29000d48c93ca.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - |
+    Updated :func:`~.from_ising` to support ``qiskit.quantum_info.SparsePauliOp``.
+    ``qubit_op`` argument can currently accept opflow operators,
+    but it has been superseded by ``SparsePauliOp``.
+    Opflow will be deprecated in a future release and subsequently removed after that.
```
woodsp-ibm, line 13: This text is similar again to the suggestion made for the from_ising note. So again is thats ok probably we want the same/similar here too.

qiskit_optimization/translators/ising.py
t-imamichi: Thanks. I include it. cc41551

qiskit_optimization/translators/ising.py
t-imamichi: Thanks. I include it. cc41551

releasenotes/notes/support-sprase-pauli-op-00d29000d48c93ca.yaml
t-imamichi: Thanks. I include it. cc41551

releasenotes/notes/support-sprase-pauli-op-00d29000d48c93ca.yaml
t-imamichi: Thanks. I include it. cc41551

qiskit_optimization/problems/quadratic_program.py
```diff
@@ -1007,13 +1009,19 @@ def substitute_variables(
...
+    ) -> Tuple[Union[OperatorBase, SparsePauliOp], float]:
         """Return the Ising Hamiltonian of this problem.
 
         Variables are mapped to qubits in the same order, i.e.,
         i-th variable is mapped to i-th qubit.
         See https://github.com/Qiskit/qiskit-terra/issues/1148 for details.
 
+        Args:
+            opflow: The output object is an OpFlow's operator if True.
+                Otherwise, it is ``SparsePauliOp``. (default: True)
```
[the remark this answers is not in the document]
t-imamichi: Thank you for your idea. I updated it to refer to `to_ising` function.

woodsp-ibm (approved):
LGTM. I see the release note has a minor spelling mistake in its name `support-sprase-pauli-op` where its `sprase` instead of `sparse` - but since the name never shows up anywhere outside of looking at the files themselves it's of little consequence.

t-imamichi:
Good catch. But, it is a bit complicated to rename reno. So, I will merge it without renaming it.

PR #475, opened 2023-03-15
Remove unnecessary `html_theme_path` option

Eric-Arellano:
See https://github.com/Qiskit/qiskit-finance/pull/244 for the motivation.

I did not test this because the docs fail to build on my M1, but as shown in https://github.com/Qiskit/qiskit-finance/pull/244, this should have no impact on the docs build.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -86,13 +86,6 @@ nbsphinx_prolog += link_str + "{{ docname }}"
 
 # -- General configuration ---------------------------------------------------
 
-# If your documentation needs a minimal Sphinx version, state it here.
-#
-# needs_sphinx = '1.0'
-
-# Add any Sphinx extension module names here, as strings. They can be
-# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
-# ones.
 extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.autodoc",
@@ -187,18 +180,7 @@ modindex_common_prefix = ["qiskit_optimization."]
 
 # -- Options for HTML output -------------------------------------------------
 
-# The theme to use for HTML and HTML Help pages.  See the documentation for
-# a list of builtin themes.
-#
-#
 html_theme = "qiskit_sphinx_theme"
-
-html_theme_path = [".", qiskit_sphinx_theme.get_html_theme_path()]
-
-# Theme options are theme-specific and customize the look and feel of a theme
-# further.  For a list of options available for each theme, see the
-# documentation.
-#
 html_theme_options = {
     "logo_only": False,
     "display_version": True,


t-imamichi:
Seems that we need to add two words "eigensolvers" and "vqeresult" in `.pylintdict`. I'm not sure how this PR affects spell check...

```
qiskit_optimization/runtime/vqe_client.py:docstring of qiskit_optimization.runtime.vqe_client.VQERuntimeResult:8: : Spell check: eigensolvers: The VQEResult class has been superseded by the qiskit.algorithms.minimum_eigensolvers.VQEResult class. This class will be deprecated in a future release and subsequently removed after that..
qiskit_optimization/runtime/vqe_client.py:docstring of qiskit_optimization.runtime.vqe_client.VQERuntimeResult:8: : Spell check: vqeresult: The VQEResult class has been superseded by the qiskit.algorithms.minimum_eigensolvers.VQEResult class. This class will be deprecated in a future release and subsequently removed after that..
```

woodsp-ibm:
> I'm not sure how this PR affects spell check...

Machine Learning has a similar problem with the same change there - its picking up on parts of longer class path from deprecation messages that it did not before. I was looking for what else changed around the spell checker. So far I have drawn a blank since a PR went through yesterday ok in ML (here its been a while and sphinxcontrib-spelling changed since the last PR run here - not so in ML so I am not sure why)

woodsp-ibm:
It does seem to be due to something very recently changing. A PR that went through OK yesterday in ML fails today with the spelling issue. See conversation in Qiskit/qiskit-machine-learning#573 for further detail.

t-imamichi:
I see this error is not related to this PR. The scheduled CI failed with the same error. I will make a fix.
https://github.com/Qiskit/qiskit-optimization/actions/runs/4432843393

t-imamichi (approved):
LGTM

Eric-Arellano:
> I don;t know about this, for sure VQResult the parent is deprecated, which kinda means something needs to happen here, I am just not sure about its message being injected - its rather confusing here I think

Oh, I see. So the subclass is not setting its own docstring and thus inherits the `__doc__` of its parent class, which is deprecated. The tactical fix here would be for the subclass to set its own docstring, rather than inheriting.

More generally, though, this is an unfortunate downside of us choosing to add the deprecation message by modifying the docstring, rather than the original approach of having a Sphinx extension only add the deprecation to the generated docs at docs generation time. I do not see a way around this problem if we want to keep changing docstring at runtime—this is how Python inheritance of docstring works, and I can't think of something clever.

If that's a dealbreaker, then we could go back to modifying docstring via a Sphinx extension. But you lose benefits that people wanted like `help()` showing the deprecation.

--

(Should we move this conversation to a GH issue? Also this PR is ready to land, this discussion is orthogonal to this PR)

woodsp-ibm:
But it does have its own docstring https://github.com/Qiskit/qiskit-optimization/blob/5a03fd4032df14626ab6c6a54029ac26f875e121/qiskit_optimization/runtime/vqe_client.py#L328-L333 thats the text in the above image that precedes the deprecation notice.

Yes this change is orthogonal to deprecation issue - it just showed here for the first time and broke CI since we do spell checking here in the apps and it introduced new text. I am ok with taking this up in another issue to let this go through but continue the conversation around deprecation.

Eric-Arellano:
> But it does have its own docstring

From Python's perspective, this is because it does not have its own docstring for `__init__`. The class docstring is a different thing, per Python's object model. So, `VQERuntimeResult.__init__()` inherits the docstring of its parent's `__init__`, w/ the deprecation message.

Now, Sphinx is fancier than that and combining the class docstring with the `__init__` docstring. But `add_deprecation_to_docstring` has nothing to do with Sphinx, since we wanted to add the deprecation via Python so it shows up in `help()`.

woodsp-ibm:
I don't know in this case then. Often the subclass will not override the docstrings. In this case with the Terra version deprecation in the Opt docs it just seems odd. Maybe we just have to add a docstring if/when we see this is happening. Had it had other text in the init there in VQEResult, beyond the args, maybe we would have already seen that here.

Eric-Arellano:
>  In this case with the Terra version deprecation in the Opt docs it just seems odd. Maybe we just have to add a docstring if/when we see this is happening. 

Yeah, I think so, unfortunately.

We could avoid this by switching from modifying docstring to instead the Sphinx extension, but some big downsides:

1. More complex deployment, including that repos must explicitly opt into the functionality by adding the extension to `conf.py`.
2. Doesn't update `help()`, which a few people wanted.

So, this quirk seems to be the "lesser sin".

woodsp-ibm:
I was wondering, VQEResult Args: aspect of its init docstring does not show up in this derived class. Is there some place it could be put e.g. Raises: where Sphinx will not bring it down. Maybe that would end up with it in not such a nice place documentation wise so it ends up being a trade-off.

woodsp-ibm (approved):
What we ended up discussing here, that was seen at the time of this change, was from unrelated decorator change in Terra. I think this cane/should be merged. How to deal with VQEClient and the fact that VQEResult is deprecated is orthogonal to this PR.

t-imamichi (approved):

PR #476, opened 2023-03-16
Add words to pylintdict

t-imamichi:
### Summary

Fix CI failure.
eigensolvers
 https://github.com/Qiskit/qiskit-optimization/actions/runs/4432843393


woodsp-ibm (approved):
I will note that one of the jobs had a failure in unit tests but passed when re-run. I did not look too much so hopefully this does not become more frequent.

PR #431, opened 2022-11-04
[WIP] Add support for Python 3.11

manoelmarques:
### Summary
Following Qiskit Terra, add support to python 3.11

### Details and comments
The following libraries don't support python 3.11:
- CPLEX

The pylint minimum version now is 2.15.0

PR #481, opened 2023-03-20
Add support for Python 3.11

t-imamichi:
### Summary

Adds support for Python 3.11.

updates tox.ini to use the stable version of aer as well as #480

supersedes #431
closes #431

### Details and comments

Co-authored-by: Manoel Marques <Manoel.Marques@ibm.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8]
+        python-version: [3.7]
     steps:
       - name: Print Concurrency Group
         env:
@@ -113,7 +113,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.9, '3.10', 3.11]
+        python-version: [3.7, 3.8, 3.9, '3.10', 3.11]
         include:
           - os: macos-latest
             python-version: 3.8
@@ -195,7 +195,7 @@ jobs:
           fi
           coverage3 combine
           mv .coverage ./ci-artifact-data/opt.dat
-        if: ${{ matrix.python-version == 3.8 }}
+        if: ${{ matrix.python-version == 3.7 }}
         shell: bash
       - uses: actions/upload-artifact@v3
         with:
@@ -222,7 +222,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, '3.10']
+        python-version: [3.7, 3.8]
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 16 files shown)


woodsp-ibm:
I think 3.7 support should not be removed until it's dropped from Qiskit Terra and end of July (0.25.0 release). We can add 3.11 support before then, but really need to wait until then to drop 3.7.

t-imamichi:
Thank you for the information. I will revert the remove of python 3.7

.github/workflows/main.yml
```diff
@@ -136,9 +136,13 @@ jobs:
...
         run: |
           source "$CONDA/etc/profile.d/conda.sh"
-          conda create -n scsenv python=${{ matrix.python-version }}
-          conda activate scsenv
-          conda install -y scs lapack cvxpy -c conda-forge
+          if [ "${{ matrix.python-version }}" == "3.11" ]; then
+            conda create -y -n scsenv python=${{ matrix.python-version }} -c conda-forge
+          else
+            conda create -y -n scsenv python=${{ matrix.python-version }}
+            conda activate scsenv
```
woodsp-ibm, line 143: I note for 3.11 the env is created but not activated? I am not sure what level of support there is for  conda and Python 3.11 - maybe this is why. I can understand packages not being available.

woodsp-ibm:
Should we have reno to state that we now support Python 3.11

woodsp-ibm:
I just noticed in looking at things here that neither the 3.10 builds, nor the new 3.11 are required to pass for CI. I can understand 3.11 as as that is new and will need to be configured. Possibly the same for 3.10 as that was changed from 3.10.6. But in looking at #467 the 3.10.6 builds are not required there either. Maybe this was an oversight - so just checking if there is some reason otherwise I think we want these required otherwise failures in these, where all else passed, would allow a PR to me merged. I imagine its unlikely, buts its possible. I should probably change things to make them required.

.github/workflows/main.yml
t-imamichi: I copied it from #431. I'm not familiar with conda and don't know the line. But, it looks strange, so I try to update it.

t-imamichi:
Good catch. Python 3.10 should be required though 3.11 is not requred due to the absense of cplex support. I look for a way to make 3.10 required.

I also write a reno.

.github/workflows/main.yml
t-imamichi: I updated to use `-c conda-forge` for `conda create` for all python versions because it is consistent with installing `scs`, `lapack`, and `cvxpy` from conda-forge.  ad22ea2

woodsp-ibm:
>  I could not find how to control "required" tag. 

Its administered via repo settings. I updated it. For the other PR #481, since this changes the workflow, I did not do anything about 3.10.6 as once that is updated after this is merged the jobs should be the same as here.

woodsp-ibm (approved):

t-imamichi:
Thanks!

PR #485, opened 2023-03-22
Upgrade Pylint to 2.16.2

t-imamichi:
### Summary

- Updates the minimum Pylint version to 2.16.2 following Terra's PR https://github.com/Qiskit/qiskit-terra/pull/9615
- Updates pylintrc file and github action workflow using #431
- Applies black to setup.py

Co-authored-by: Manoel Marques <Manoel.Marques@ibm.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -166,15 +166,6 @@ jobs:
           make mypy
         if: ${{ !cancelled() }}
         shell: bash
-      - name: Run lint latest version
-        run: |
-          if [ "${{ matrix.os }}" == "windows-2019" ]; then
-            source "$CONDA/etc/profile.d/conda.sh"
-            conda activate scsenv
-          fi
-          pip install -U -r requirements-dev.txt
-          make lint
-        shell: bash
       - name: Optimization Unit Tests under Python ${{ matrix.python-version }}
         uses: ./.github/actions/run-tests
         with:
--- a/.pylintrc
+++ b/.pylintrc
@@ -58,8 +58,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=no-self-use,        # disabled as it is too verbose
-    fixme,              # disabled as TODOs would show up as warnings
+disable=fixme,              # disabled as TODOs would show up as warnings
     protected-access,   # disabled as we don't follow the public vs private
                         # convention strictly
     duplicate-code,     # disabled as it is too verbose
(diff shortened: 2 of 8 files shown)


woodsp-ibm:
A general comment here. For the apps, given its a lot less code, we used to lint on the latest version as well as whatever Terra used. Yes when it changed it might affect/break CI. I am not fan of fixing a specific version, and linting against something that may be older, but it is a tradeoff I agree. Will we end up updating it or forget it. Already pylint is at 2.17.1. Given the apps are not as active now maybe its something we have to do or need to do

t-imamichi:
I copied the removal of the latest lint from #431. It makes sense to me to apply the same pylint version as Terra because applying the latest pylint version has a risk to break CI suddenly due to breaking change of pylint.

woodsp-ibm:
As far as I recall the app repos were set to lint using the same of lint as Terra as well as the new version. Its a tradeoff as I mentioned - I did not like being "stuck" on some on older version and not taking advantage of any new features/capability that might help with code quality, hence we did both. Yes it sometimes fails due to new checks that were not done before. For #431 the older versions did not work and hence Manoel had to remove that

t-imamichi:
As for mypy, we always apply the latest one because we don't pin mypy as follows.
As for pylint, it might be enough to set `pylint>=2.16.2` and apply only the latest one in the CI.
https://github.com/Qiskit/qiskit-optimization/blob/8044ee0e74c7560908d30328dfcaf430a7ad4bf6/requirements-dev.txt#L17

t-imamichi:
I updated requirements-dev.txt to apply the latest pylint at CI 5d7f3e4. It looks working.

woodsp-ibm:
Ok, my rational to suggest pinning them and having another job running on the latest which was not required to pass CI and hence would not block things if it failed was more a response to this

>the latest pylint version has a risk to break CI suddenly

We have usually just done emergency fixes to sort things, for both this and mypy since that can break CI if/when new rules  come in. Anyway its something we can consider going forwards i.e. have main CI running on pinned versions we know work, and an extra job testing against latest versions such that we can fix things more leisurely if they break.

woodsp-ibm:
Ok, my rational to suggest pinning them and having another job running on the latest which was not required to pass CI and hence would not block things if it failed was more a response to this

>the latest pylint version has a risk to break CI suddenly

We have usually just done emergency fixes to sort things, for both this and mypy since that can break CI if/when new rules  come in. Anyway its something we can consider going forwards i.e. have main CI running on pinned versions we know work, and an extra job testing against latest versions such that we can fix things more leisurely if they break.

woodsp-ibm (approved):
I must admit the extra brackets in the tests, for me improved readability and more clearly signaled the intent.

t-imamichi:
Thanks. Since the current CI tests with the latest pylint, this PR does not increase the risk of CI breakage. Let's discuss pinning pylint version if necessary.

PR #489, opened 2023-04-05
Various docs changes for CI

woodsp-ibm:
### Summary

See Qiskit/qiskit-nature#1109 for more detail.

Basically does the same three things

- Removes dependency on sphinx-autodoc-typehints
- Adds default thumbnail config as that changed in nbsphinx 0.9
- Updates tox.ini gettext section used for translatable strings due to ipykernel not being installed by default now

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -58,6 +58,7 @@ deps =
   -r{toxinidir}/requirements.txt
   -r{toxinidir}/requirements-dev.txt
   sphinx-intl
+  jupyter
 commands =
   sphinx-build -W -T --keep-going -b gettext docs/ docs/_build/gettext {posargs}
   sphinx-intl -c docs/conf.py update -p docs/_build/gettext -l en -d docs/locale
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -58,6 +58,7 @@ deps =
   -r{toxinidir}/requirements.txt
   -r{toxinidir}/requirements-dev.txt
   sphinx-intl
+  jupyter
```
t-imamichi, line 61: Is it necessary? I succeeded to run `tox -egettext` without this line and got the following message.
```
  gettext: OK (97.34=setup[60.66]+cmd[34.39,2.28] seconds)
  congratulations :) (97.44 seconds)
```

t-imamichi: Thank you for the details. Let's merge them as they are.

t-imamichi (approved):

t-imamichi:
Is this backport stable?

woodsp-ibm:
> Is this backport stable?

We can do, I have done that in other places. Since the next release may be a little way off yet its probably better to do so.

PR #492, opened 2023-04-10
Tag cell with nbsphinx-thumbnail

woodsp-ibm:
### Summary

After update the conf.py, like the other repos, to list the default thumbnail image. since with nbsphinx 0.9 it no longer recognizes the older way of doing that with the png file, some tiles "lost" their images, as below. Only the 5 with explicit images in the conf.py show up

image

This is how it looked before without the conf/py default thumbnail change where the tile that should have the sphere image does not.

image

### Details and comments

It seems that the last image in the notebook was automatically selected how things were before. I added an explicit `nbsphinx-thumbnail` tag to the respective cell (as we have done in other repos which is why I did not see that there). I also selected a better image, I think, for the qaoa runtime. This is how it looks now.

image

### DIFF
--- a/docs/tutorials/06_examples_max_cut_and_tsp.ipynb
+++ b/docs/tutorials/06_examples_max_cut_and_tsp.ipynb
@@ -927,7 +927,11 @@
   {
    "cell_type": "code",
    "execution_count": 21,
-   "metadata": {},
+   "metadata": {
+    "tags": [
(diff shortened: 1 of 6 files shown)


woodsp-ibm:
We can probably StableBackport this so the docs out there can be updated so the images return to those that did have it before.

t-imamichi (approved):

PR #497, opened 2023-04-25
Bump python version for CI check

t-imamichi:
### Summary

Bump python version for CI check, deprecation, and coverage.

### DIFF
--- a/.github/workflows/deploy-code.yml
+++ b/.github/workflows/deploy-code.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: [3.7]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-code.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7]
+        python-version: [3.8]
```
woodsp-ibm, line 25: I am not sure about changing this at present. I do not think it really matters but since we still will be producing releases for 3.7, including this imminent one, maybe its better to leave it at 3.7 until we drop that and 3.8 becomes the min level we support.

woodsp-ibm:
Given this change the branch rules for main will need to be adjusted to drop the 3.7 checks and deprecation ones from being required and make this the new 3.8 ones instead. I can do this if you like.

.github/workflows/deploy-code.yml
t-imamichi: Good point. I should revert this.

t-imamichi:
Thanks. Could you adjust them?

woodsp-ibm (approved):

PR #498, opened 2023-04-25
Update legacy unit tests

t-imamichi:
### Summary

Update some legacy tests by changing `PendingDeprecationWarning` with `DeprecationWarning` due to deprecation opflow.


woodsp-ibm (approved):
LGTM. I note there are quite a few deprecations messages coming out from the tests that arguably would be nice to filter out and/or ensure we are are using opflow when we have to test old paths and not for convenience of building something out in a test (not sure if thats the case here as I did not look in detail)

t-imamichi:
Yes, the legacy tests intentionally use opflow and contain a lot of deprecation warnings. I will investigate how to mitigate deprecation messages in another PR.

PR #504, opened 2023-05-08
Fix language toggle for docs

Eric-Arellano:
### Summary

The language selection was broken because:

1. The extension should set `language_label`, not `version_label`
2. Sphinx 6 stopped activating jQuery by default.

This also makes some other improvements:

* Fixes the docs build on macOS. cplex cannot be installed on Apple Silicon.
* Removes outdated HTML templates.
* Pins the theme to 1.11 for better stability. I will update to new theme releases when ready.


tox.ini
```diff
@@ -47,8 +47,13 @@ commands =
   coverage3 report
 
 [testenv:docs]
+# Editable mode breaks macOS: https://github.com/sphinx-doc/sphinx/issues/10943
+usedevelop = False
 envdir = .tox/docs
 basepython = python3
+deps =
+  -r{toxinidir}/requirements-dev.txt
+  .
```
Eric-Arellano, line 56: I'm not really sure why the `.` was necessary - I thought Sphinx normally would install that for us. It means to install optimization itself.

t-imamichi (approved):
Thanks!

Eric-Arellano:
Thanks! Ditto that I can't hit merge on this

PR #505, opened 2023-05-08
Turn off CI for forks

Eric-Arellano:
Most people agree that GitHub made a mistake by having forks run CI both on the fork and in the upstream repository. It wastes computing resources (which has a real environmental cost!).

If people really want to run CI in their fork, they can always modify their fork's config to revert this change.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,6 +31,7 @@ concurrency:
 
 jobs:
   Checks:
+    if: github.repository_owner == 'Qiskit'
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


t-imamichi (approved):
LGTM. Thanks.

Eric-Arellano:
Thanks for the review! I don't have write access, if you could please hit merge.

Eric-Arellano:
Github still is dealing with an outage: https://www.githubstatus.com

woodsp-ibm:
@t-imamichi I responded back - the mergify.yml file needs updating - the job it used to check success was the 3.7 version which you changed to 3.8 so that now needs to be the 3.8 one (two places in that file)

t-imamichi:
Thank you, @woodsp-ibm. I have merged #507.

PR #507, opened 2023-05-09
Update mergify config

woodsp-ibm:
### Summary

#497 updated the `Deprecation_Messages_and_Coverage` job of the main workflow to use Python 3.8 (was 3.7 before). The outcome of this is used by mergify for its readiness check. This updates the mergify config to match so automerge label will have an effect again.

This PR likely needs to be manually merged though since its changing the config.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,12 +1,12 @@
 queue_rules:
   - name: automerge
     conditions:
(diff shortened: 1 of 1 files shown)


t-imamichi (approved):
Thanks!

PR #510, opened 2023-05-12
Remove unnecessary CSS files

Eric-Arellano:
No one was using the custom directives `galleryitem`, `customgalleryitem`, `includenodoc`, `customcarditem`. 

After removing this, it's safe to remove `sphinx-gallery` because it's not used. That allows removing `gallery.css`, which is meant to override `sphinx-gallery`'s default styling.

This also removes two CSS files:

* `custom.css`: There is no `.toggle` element anymore
* `style.css`: its 3 rules are holdovers from the Read The Docs theme

`CustomCalloutItemDirective` is used though. It's used in a few other projects too, so I'm likely going to move this to `qiskit_sphinx_theme` in its 1.13 release so that you don't have to copy and paste the code. Tracked by https://github.com/Qiskit/qiskit_sphinx_theme/issues/323.


Eric-Arellano:
Oops, I didn't realize this project was also using custom directives. Updated.

t-imamichi (approved):

PR #521, opened 2023-05-26
Update tutorial 11 intro text around pip installed free solvers

### LINKED ISSUE #488: Tutorial 11_using_classical_optimization_solvers_and_models and latest cplex
This is run at present on Python 3.8. The latest cplex https://pypi.org/project/cplex/22.1.1.0/ only has wheels for 3.9 and 3.10. As such we will likely need to update CI for what Python we produce the docs/tutorials with, if we want to run with this. At present, since its running on 3.8, CI ends up pulling in an older version that still has these wheels. 

Also this text maybe can  be revised

> Note that these solvers installed via pip are free versions, which have some limitations such as number of variables. Also note that the latest version 20.1 of CPLEX is available only for Python 3.7 and 3.8 as of July 2021. See the following pages for details.

Perhaps we can write it in a way that is more timeless so as to avoid this getting stale.

woodsp-ibm:
### Summary

Fixes #488

I removed the reference to cplex version totally to avoid having something that would get outdated and stale again. I think the note and the links to the respective packages, for further details, suffices.

### DIFF
--- a/docs/tutorials/11_using_classical_optimization_solvers_and_models.ipynb
+++ b/docs/tutorials/11_using_classical_optimization_solvers_and_models.ipynb
@@ -19,7 +19,7 @@
     "\n",
     "If you want to use Gurobi and Gurobipy, you need to install `pip install 'qiskit-optimization[gurobi]'`.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


t-imamichi (approved):
LGTM

PR #529, opened 2023-06-07
Update urls following repo move

woodsp-ibm:
### Summary

Following the recent move of this repo from `Qiskit` organization to `qiskit-community` this updates all the urls to the new location. The file `theme-variables.jinja` is quite out of date and being unused the conclusion was to remove this.


setup.py
```diff
@@ -48,7 +48,7 @@
     description="Qiskit Optimization: A library of quantum computing optimizations",
```
woodsp-ibm, line 48: @t-imamichi Do we want to change this at the same time. As it sounds its more like a set of optimizations you might apply to your quantum computing code. I was thinking perhaps more like

> Qiskit Optimization: A library for Optimization applications using quantum computing

Nature and ML look like this

> Qiskit Nature: A library of quantum computing chemistry/physics experiments
> Qiskit Machine Learning: A library of quantum computing machine learning experiments

So in that same vein it would be 

> Qiskit Optimization: A library of quantum computing optimization experiments.

although I guess for me `quantum computing optimization` could be misinterpreted easily

t-imamichi: Thank your for your suggestion. I like the first one

> Qiskit Optimization: A library for optimization applications using quantum computing

t-imamichi (approved):
LGTM

PR #537, opened 2023-08-21
Replace qiskit.algorithms with qiskit_algorithms

t-imamichi:
### Summary

Replace `qiskit.algorithms` with `qiskit_algorithms` except `QAOAClient` and `VQEClient` because they are deprecated.

This PR requires qiskit-algorithms>=0.2.0 to switch to `qiskit_algorithms.utils.algorithm_globals`.

### Details and comments

- [x] Since qiskit-algorithms does not support python 3.7, unit tests fail with Python 3.7. I made #538 to drop Python 3.7 support.
- [x] #542


woodsp-ibm:
* We should add `qiskit_algorithms` to the requirements file.

* Once 0.2.0 of qiskit_algorithms in released we can also switch here `from qiskit.utils import algorithm_globals` over to the equivalent class in qiskit_algorithms i.e.  `from qiskit_algorithms.utils import algorithm_globals` and make that the min version in the requirements file. Similar to noted here /qiskit-community/qiskit-finance#291 qiskst_algorithms works with either but that file in utils in Qiskit, along with another that's only used by algorithms. I expect to be deprecated and removed when the algorithms are removed from Qiskit so its better to switch now.

t-imamichi:
Yes, I added qiskit-algorithms in requirements.txt.

OK. I will switch to `from qiskit_algorithms.utils import algorithm_globals`. When is qiskit-algorithms 0.2.0 planned? Is it soon?

woodsp-ibm:
> When is qiskit-algorithms 0.2.0 planned? Is it soon?

Should be very soon now.

t-imamichi:
It's ready to review because qiskit-algorithms 0.2.0 was released on Aug 24.

woodsp-ibm:
@t-imamichi My take is that I think the logic ought to be able to take a qiskit.algorithms instance or a qiskit_algorithms one. The instance checks there I think will prevent the former. That may end up being a tradeoff in what sort of upfront checks can be made and how much we want to switch to just using qiskit_algorithms and no longer refer to qiskit.algorithms at this point. Did you see the post in the thread in the internal algorithms related discussion channel - where you mentioned this PR it was a response to that - it includes some thoughts about this and more.

t-imamichi:
@woodsp-ibm Thanks. I tried to remove all imports of `qiskit.algorithms` 63474dc. It would work fine with `qiskit.algorithms` implicitly. But, it requires #542 to pass the unit tests because some legacy tests uses `qiskit.algorithms` that need to be removed first.

README.md
```diff
@@ -68,10 +68,11 @@ from docplex.mp.model import Model
 from qiskit_optimization.algorithms import MinimumEigenOptimizer
 from qiskit_optimization.translators import from_docplex_mp
 
-from qiskit.utils import algorithm_globals
 from qiskit.primitives import Sampler
-from qiskit.algorithms.minimum_eigensolvers import QAOA
-from qiskit.algorithms.optimizers import SPSA
+
+from qiskit_algorithms.utils import algorithm_globals
+from qiskit_algorithms.minimum_eigensolvers import QAOA
```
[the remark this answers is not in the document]
t-imamichi: Thanks. I will update it to use the short imports.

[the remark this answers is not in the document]
t-imamichi: done

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
```diff
@@ -232,6 +218,12 @@ def _solve_internal(
             eigen_result = self._min_eigen_solver.compute_minimum_eigenvalue(operator)
             # analyze results
             raw_samples = None
+            if not hasattr(eigen_result, "eigenstate"):
+                raise QiskitOptimizationError(
+                    "MinimumEigenOptimizer does not support this VQE "
+                    f"{type(self._min_eigen_solver)}. "
+                    "You can use qiskit_algorithms.SamplingVQE instead."
+                )
```
t-imamichi, line 226: Instead of early type check of `SamplingVQE` or `NumPyMinimumEigensolver`, I added a result check when `solve` is invoked.

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
```diff
@@ -232,6 +218,12 @@ def _solve_internal(
             eigen_result = self._min_eigen_solver.compute_minimum_eigenvalue(operator)
             # analyze results
             raw_samples = None
+            if not hasattr(eigen_result, "eigenstate"):
+                raise QiskitOptimizationError(
+                    "MinimumEigenOptimizer does not support this minimum eigensolver "
+                    f"{type(self._min_eigen_solver)}. "
+                    "You can use qiskit_algorithms.SamplingMinimumEigensolver instead."
+                )
```
t-imamichi, line 226: Instead of early type check of SamplingMinimumEigensolver or NumPyMinimumEigensolver, I added a result check when solve is invoked.

qiskit_optimization/__init__.py
```diff
@@ -26,16 +26,17 @@
 A uniform interface as well as automatic conversion between different problem representations
 allows users to solve problems using a large set of algorithms, from variational quantum algorithms,
 such as the Quantum Approximate Optimization Algorithm
-(:class:`~qiskit.algorithms.minimum_eigensolver.QAOA`), to
+(:class:`~qiskit_algorithms.QAOA`), to
 `Grover Adaptive Search <https://arxiv.org/abs/quant-ph/9607014>`_
 (:class:`~algorithms.GroverOptimizer`), leveraging
-fundamental :mod:`~qiskit.algorithms.minimum_eigensolver` provided by Qiskit Terra.
+fundamental :mod:`~qiskit_algorithms.minimum_eigensolvers` provided by
```
woodsp-ibm, line 32: With qiskit_algorithms this will not form a link as it will not resolve to any documentation since the minimum_eigensolvers module is not documented as it was before (before it was not a link either since it lacked the `s` at the end). I think its ok.

The Qiskit Algorithms link on the next line I think would be better to the published docs though so you go from docs to docs ie. use https://qiskit.org/ecosystem/algorithms/

And for the module above if you preferred it to be a link then it could be changed to make it a link to where the `Minimum Eigensolvers` are documented ie this https://qiskit.org/ecosystem/algorithms/apidocs/qiskit_algorithms.html#minimum-eigensolvers

qiskit_optimization/algorithms/minimum_eigen_optimizer.py
woodsp-ibm: It will certainly result in a more friendly error message that is easier to understand. This is of course detected though after its run compute_minimum_eigenvalue which may take a lot of time/resources that ends up being wasted. An IDE via the typehinting should flag VQE (which I think we we most concerned about in 0.5 since in the past you could use this) as not being valid to be used for this algorithm. I think its ok - we can see how it goes - one suggestion I had made was to get the class name string from the solver instance and compare that up front to see if it was one of the primitive based VQE's we know about ie - that being just a string check does not require anything to be imported. But we could always add this later if users are still tripping up on VQE - since 0.5 has been out a while hopefully users are familiar enough now. And its quite clear now in the docs what is supported...

image

releasenotes/notes/switch-to-qiskit-algorithms-b0af38fe547bf5b5.yaml
```diff
@@ -0,0 +1,24 @@
+---
+prelude: >
+    Qiskit Optimization 0.6 switches from ``qiskit.algorithms`` of Qiskit Terra
+    to `Qiskit Algorithms <https://github.com/qiskit-community/qiskit-algorithms>`_.
+    Qiskit Optimization 0.6 drops supports of the former algorithms based on
```
woodsp-ibm, line 5: Although support for the qiskit.algorithms primitive based ones is dropped, for the moment since they are compatible with what was migrated to qiskit_algorithms they should still work so any users code should still work as they transition/upgrade it - right.

test/algorithms/test_grover_optimizer.py
```diff
@@ -40,6 +41,7 @@
 class TestGroverOptimizer(QiskitOptimizationTestCase):
     """GroverOptimizer tests."""
 
+    @unittest.skipUnless(optionals.HAS_AER, "qiskit-aer is required to run this test")
```
woodsp-ibm, line 44: At some point, since the reference Sampler is probably fast enough, we could probably change tests like these so we do not require Aer any longer and this test no longer is an optional one - though CI of course will install Aer. I would leave things as they are now. Its more just a comment.

qiskit_optimization/__init__.py
t-imamichi: Thanks. I updated it to mention minimum eigensolvers with https://qiskit.org/ecosystem/algorithms/apidocs/qiskit_algorithms.html#minimum-eigensolvers instead of `~qiskit_algorithms.minimum_eigensolvers`

releasenotes/notes/switch-to-qiskit-algorithms-b0af38fe547bf5b5.yaml
t-imamichi: I'm wondering whether I have to mention qiskit.algorithms is not officially supported by qiskit-optimization but it works actually.

test/algorithms/test_grover_optimizer.py
t-imamichi: Sorry that I removed this line accidentally in #542. I intended to revert the removal in this PR. It requires aer sampler because terra sampler takes much more time than aer sampler (#451 for details).

releasenotes/notes/switch-to-qiskit-algorithms-b0af38fe547bf5b5.yaml
woodsp-ibm: The intent was really to have it work so users had a easier transition. Code that they had working based on primitives should still work with this new version. Then they can simply switch the code to qiskit_algorithms and it should work just like before.

qiskit_optimization/__init__.py
t-imamichi: Sorry. I forgot to push it. 10d057b

woodsp-ibm (approved):

svdbtno:
@t-imamichi When can we expect this to be released?

PR #533, opened 2023-07-14
Remove Cplex 3.11 restriction now that Cplex supports it

woodsp-ibm:
### Summary

There was a new [Cplex release 22.1.1.1](https://pypi.org/project/cplex/22.1.1.1/#files) which added wheels for Python 3.11. This PR simply removes the version restriction from setup.py. Since the 3,11 support overall of Qiskit Optimization has not yet been released this did not need any reno, but I did update the reno from the 3.11 PR to remove the note above CPLEX not supporting 3.11.

PR #538, opened 2023-08-21
Drop python 3.7 support

t-imamichi:
### Summary

Qiskit has dropped support Python 3.7 and this follows suit for Qiskit Optimization.
follows https://github.com/qiskit-community/qiskit-finance/pull/288

depends on #536

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -116,7 +116,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.7, 3.8, 3.9, '3.10', 3.11]
+        python-version: [3.8, 3.9, '3.10', 3.11]
         include:
           - os: macos-latest
             python-version: 3.8
@@ -222,7 +222,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.7, 3.8]
+        python-version: [3.8, '3.10']
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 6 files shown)


woodsp-ibm:
pyproject.toml and tox.ini both have `py37` entries that should be removed too

.github/workflows/main.yml
```diff
@@ -222,7 +222,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.7, 3.8]
+        python-version: [3.8, '3.10']
```
woodsp-ibm, line 225: Maybe the 3.10 instead of 3.11 is due to dependents in the tutorials not being available? Cplex is now available #533 - so maybe once thats merged? Is there anything else  - that would have these done at the lowest and highest supported versions.

t-imamichi: Now that I merge #533, I will update the line to use 3.11 instead of 3.10.

t-imamichi:
@woodsp-ibm I guess we need to configure somewhere to update the following entries to Python 3.8. Do you have any ideas?
image

woodsp-ibm:
>  I guess we need to configure somewhere to update the following entries to Python 3.8. Do you have any ideas? We can update the configuration after merging this PR.

They are edited under the Settings for the repo - select Branches from the left and you will see Branch Protection Rules and you can Edit the rules for the main branch. Should probably make sure the new 3.11 tutorials is required. Once you edit the rules the checks above should alter accordingly. 

Don't forget also the comment above about the two other files that should be updated - i.e the removal of py37 in them now that this drops 3.7 support etc.

woodsp-ibm:
Looks good, just missing these two file being updated to remove `py37` envs

https://github.com/qiskit-community/qiskit-optimization/blob/3f21208d1a9fa448c7fc602e3ff527d1b601ba8c/tox.ini#L3

https://github.com/qiskit-community/qiskit-optimization/blob/3f21208d1a9fa448c7fc602e3ff527d1b601ba8c/pyproject.toml#L7

And I note in the latter (pyproject.tonl) we forgot to add py311 to that list for black so maybe that can be added, and we can fix it here, while removing the py37

t-imamichi:
Thanks. I updated tox.ini and pyproject.toml.

woodsp-ibm (approved):
LGTM

PR #542, opened 2023-08-30
Remove runtime program, quantum instance, and opflow

t-imamichi:
### Summary

This PR removes runtime clients, which have been already removed at IBM Quantum Platform, and `QuantumInstance`-based algorithms.

- [x] remove `QuantumInstance`-based legacy algorithms
- [x] remove VQEClient, QAOAClient, and VQERuntimeResult
- [x] remove Opflow
- [x] update tutorials
- [x] reno

### Details and comments

~~It still uses opflow-based `PauliSumOp` somewhere for the compatibility. E.g., `to_ising` function to return `PauliSumOp` with the default option (option `opflow=False` returns `SparsePauliOp`). Should I remove them all as well?~~
→ Opflow is also removed.


woodsp-ibm:
>It still uses opflow-based PauliSumOp somewhere for the compatibility. E.g., to_ising function to return PauliSumOp with the default option (option opflow=False returns SparsePauliOp). Should I remove them all as well?

Yes, the intent is to cease use of deprecated function like that so supporting just SparsePauliOp is all that should be done.

In terms of other deprecated code in Opt, I did a quick search and this was all that came up for me. We should remove this too - in this case its time is up anyway (could have been removed last release)

https://github.com/qiskit-community/qiskit-optimization/blob/e2855f2b6337a76da41fcab15b188d9904642f0f/qiskit_optimization/applications/graph_optimization_application.py#L83-L89

docs/migration/01_migration_guide_to_v0.5.ipynb
```diff
@@ -438,7 +414,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "`GroverOptimizer` supports both `QuantumInstance` and `BaseSampler`. But users must specify one of them."
+    "`GroverOptimizer` supports `BaseSampler` instead of `QuantumInstance`."
```
woodsp-ibm, line 417: I imagine this is needed because we still run this. I think what we did in Nature (or Max more correctly did) was to export it as a .rst so the content could stay the same, and still applied to migrating to the stated version, but it was no longer run.

releasenotes/notes/remove-quantum-instance-and-runtime-2c00393c0de1b9a9.yaml
```diff
@@ -0,0 +1,7 @@
+---
+upgrade:
+  - |
+    Support for ``QuantumInstance``-based algorithms has been removed.
```
woodsp-ibm, line 4: Perhaps we might like to say that only primitive based algorithms are supported now and link to the 0.5 migration guide. Of course beyond that migrating to qiskit_algorithms too but I guess we'll cover that elsewhere

releasenotes/notes/remove-quantum-instance-and-runtime-2c00393c0de1b9a9.yaml
```diff
@@ -0,0 +1,7 @@
...
+  - |
+    Classes ``VQEClient`` and ``QAOAClient`` have been removed.
```
woodsp-ibm, line 6: Perhaps we should say that using primitives under a Session with the Runtime is the "replacement" function for these.

docs/migration/01_migration_guide_to_v0.5.ipynb
t-imamichi: I got it. Since it's a migration guide for 0.5, I should not change it. I replace it with rst file.

t-imamichi:
I removed opflow and updated tutorials and reno following Steve's feedback.

qiskit_optimization/algorithms/grover_optimizer.py
```diff
@@ -185,11 +139,8 @@ def solve(self, problem: QuadraticProgram) -> OptimizationResult:
             AttributeError: If the quantum instance has not been set.
             QiskitOptimizationError: If the problem is incompatible with the optimizer.
         """
-        if self._sampler is None and self._quantum_instance is None:
-            raise ValueError("A quantum instance or sampler must be provided.")
-
-        if self._quantum_instance is not None and self._sampler is not None:
-            raise ValueError("Only one of quantum_instance or sampler can be passed, not both!")
+        if self._sampler is None:
+            raise ValueError("A sampler must be provided.")
```
woodsp-ibm, line 143: This is just a comment on this. Given that a Sampler is now the only option with the removal of QI we could have defaulted the Sampler to the qiskit.primitives one if it was None in the constructor (and noted this for the sampler param there). That way the default value does not lead to an error - which seems strange. I get why it was done having to support both QI and Sampler. Since that is more of an improvement, and the GroverOptimizer was being looked at to be improved anyway, I think we can live with it like this, so as I said just a comment.

woodsp-ibm (approved):
Seems good to me. I left a comment above but its just that a comment - I don't see it blocking this.

qiskit_optimization/algorithms/grover_optimizer.py
t-imamichi: Yes, we can make Sampler mandatory. But we need to change the argument order (need to move `sampler` first or second). Do you think whether we can change the argument order of `GroverOptimizer.__init__`?

PR #543, opened 2023-08-31
Switch from qiskit-terra to qiskit

woodsp-ibm:
### Summary

Similar to qiskit-community/qiskit-machine-learning#680

Switches over to `qiskit` from `qiskit-terra` 

* requirements.txt updated
* CI updated for install main dependencies
* CONTRIBUTING.MD links etc updated. Plus I removed/simplified some bits which are no longer correct and may have led to confusion down the line.

There was no Neko integration/workflow here to be removed (unlike ML)

I also updated a couple of places that referred to Terra (I was searching for Terra) in all the files. I guess I could have commented in #537 and had it done there, since this is now referring to qiskit_algorithms as the source of algorithms not Terra and while it had changed one occurrence in the main init file, I did not see it touching these files and figured I would just do it here since its really only a single line change in each.

### Details and comments

I am starting here, unlike with ML, with the setting here for the default stable install. The changes to install from source I checked out on ML so they work - we could check it out here too if needed just in case I made a slip. I compared the changes and things look comparable so it all seems fine.


t-imamichi (approved):

woodsp-ibm:
I labelled this stable backport by mistake!!! I had intended to do the master->main switch one for the translatable strings. I closed the backported PR with a similar comment.

PR #559, opened 2023-09-18
Fix stestr and mypy issues

t-imamichi:
### Summary

~~Fix "stestr run" error by pinning python-subunit version.~~
Update stestr version that has been just released.
https://github.com/mtreinish/stestr/releases/tag/4.1.0

Fix mypy by modifying colormap access.

I found these errors in the recent scheduled CI.
https://github.com/qiskit-community/qiskit-optimization/actions/runs/6217275395

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,7 +3,7 @@ matplotlib>=2.1
 black[jupyter]~=22.0
 pylint>=2.16.2
 pylatexenc>=1.4
-stestr>=2.0.0
+stestr>=4.1.0
 ddt>=1.2.0,!=1.4.0
 reno>=3.4.0
 Sphinx>=5.0
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,3 +1,2 @@
 numpy>=1.20.0
 ipython<8.13;python_version<'3.9'
-python-subunit<1.4.3
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,2 +1,3 @@
 numpy>=1.20.0
 ipython<8.13;python_version<'3.9'
+python-subunit<1.4.3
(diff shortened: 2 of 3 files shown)


requirements-dev.txt
```diff
@@ -3,7 +3,7 @@ matplotlib>=2.1
 black[jupyter]~=22.0
 pylint>=2.16.2
 pylatexenc>=1.4
-stestr>=2.0.0
+stestr>=4.1.0
```
woodsp-ibm, line 6: I guess this is fine - the constraint change that Qiskit did was simply closed. CI pulls in the latest version anyway even with >=2.0.0, but if you had some local older stestr install working this requires things to be updated if you pip install this dev file. Qiskit left their requirements-dev as it was before. I guess if you had things working locally and somehow   just updated python-subunit then it would faili if you did not have anything installed then as of now, like CI, it would install 4.1.0 and work.

woodsp-ibm (approved):
I put a comment in there - I think its ok like this but I'll leave it to you.

requirements-dev.txt
t-imamichi: Yes, I updated the version because someone may update only python-subunit and have the stestr trouble.

woodsp-ibm:
@t-imamichi I was looking through the combined deprecation messages from the build, as this was the latest. I noticed this which seems related to what you fixed here but elsewhere

> /home/runner/work/qiskit-optimization/qiskit-optimization/qiskit_optimization/applications/bin_packing.py:121: MatplotlibDeprecationWarning: The get_cmap function was deprecated in Matplotlib 3.7 and will be removed two minor releases later. Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap(obj)`` instead.

t-imamichi:
Thanks for the report. I found a line that uses `get_cmap`. I will fix it.
https://github.com/qiskit-community/qiskit-optimization/blob/e4804afef36850686df5e397bdf9b388393054d4/qiskit_optimization/applications/bin_packing.py#L121

PR #562, opened 2023-09-28
Updates to new Ecosystem theme.

arnaucasau:
### Summary
Updates to new Ecosystem theme.

Total file size increased (4.76 MB → 5.36 MB) 
index.html size increased (15 KB → 40.2 KB)
Build time increased (47s → 52s)

For more information: https://github.com/Qiskit/qiskit_sphinx_theme/issues/232

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -176,19 +176,8 @@ modindex_common_prefix = ["qiskit_optimization."]
 
 # -- Options for HTML output -------------------------------------------------
 
-html_theme = "qiskit_sphinx_theme"
-html_theme_options = {
-    "logo_only": False,
-    "display_version": True,
-    "prev_next_buttons_location": "bottom",
-    "style_external_links": False,
-    # Toc options
-    "collapse_navigation": True,
-    "sticky_navigation": True,
-    "navigation_depth": 4,
-    "includehidden": True,
-    "titles_only": False,
-}
+html_theme = "qiskit-ecosystem"
+html_title = f"{project} {release}"
 
 intersphinx_mapping = {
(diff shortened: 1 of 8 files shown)


woodsp-ibm:
In Qiskit Nature I note that you changed SubModule heading level - presumably so SubModule does not come out in left navigation sidebar. Does that need doing here - e.g. this one https://github.com/qiskit-community/qiskit-optimization/blob/cddfbbbaf529a8162e9a3180b31b38d7190650e5/qiskit_optimization/algorithms/__init__.py#L65-L66

arnaucasau:
Yes, that's right, thank you for your comment Steve! I have also noticed that happening with different entries as well (without changing the heading level there are multiple entries in the menu to the same page). I fixed it in this last commit, let me know what you think, and I will do the same with qiskit-machine-learning, and check again qiskit-nature if you want, of course.

t-imamichi (approved):

Eric-Arellano:
Thanks @t-imamichi! Arnau doesn't have write access, so this will need a merge from a maintainer.

woodsp-ibm (approved):
In the end I guess it comes down to what people/users expect - if things are consistent this way across the various projects I guess its fine. It does flatten things out and loses some of the structure this way. In qiskit-algorithms the algorithms main page has several sections for different types which done this way they will all appear in a single list - though the list is in the same order so if you know more what you are looking for. I guess there is a tension though with the right sidebar that shows that level of structure but not the detail under it (that which is flattened out in the left hand side). Lets go with this and have things done consistent like this - it seems the heading level changes are simple enough to revisited if it seems desired in the future.

PR #563, opened 2023-10-11
Move methods into class pages for docs

arnaucasau:
### Summary
Moved all the methods and attributes into class pages to speed up the building process of the documentation with the new ecosystem sphinx theme. This change helps with this [PR](https://github.com/qiskit-community/qiskit-optimization/pull/562).

See https://github.com/Qiskit/qiskit/pull/10455 for more information.

### DIFF
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -18,11 +18,9 @@
 
    .. rubric:: Attributes
 
-   .. autosummary::
-      :toctree: ../stubs/
    {% for item in all_attributes %}
       {%- if not item.startswith('_') %}
-          ~{{ name }}.{{ item }}
+   .. autoattribute:: {{ name }}.{{ item }}
       {%- endif -%}
    {%- endfor %}
    {% endif %}
@@ -33,11 +31,9 @@
 
    .. rubric:: Methods
 
-   .. autosummary::
-      :toctree: ../stubs/
    {% for item in all_methods %}
       {%- if not item.startswith('_') %}
-          ~{{ name }}.{{ item }}
+   .. automethod:: {{ name }}.{{ item }}
       {%- endif -%}
    {%- endfor %}
 
--- a/docs/_templates/autosummary/class_no_inherited_members.rst
+++ b/docs/_templates/autosummary/class_no_inherited_members.rst
@@ -18,12 +18,10 @@
 
    .. rubric:: Attributes
 
-   .. autosummary::
-      :toctree: ../stubs/
    {% for item in all_attributes %}
       {%- if item not in inherited_members %}
         {%- if not item.startswith('_') %}
-            ~{{ name }}.{{ item }}
+   .. autoattribute:: {{ name }}.{{ item }}
         {%- endif -%}
       {%- endif %}
    {%- endfor %}
@@ -35,12 +33,10 @@
 
    .. rubric:: Methods
 
-   .. autosummary::
-      :toctree: ../stubs/
    {% for item in all_methods %}
       {%- if item not in inherited_members %}
         {%- if not item.startswith('_') %}
-            ~{{ name }}.{{ item }}
+   .. automethod:: {{ name }}.{{ item }}
         {%- endif -%}
       {%- endif %}
    {%- endfor %}


woodsp-ibm:
Now #562 is not yet merged - was the intent to do this first. I noted you opened a similar PR to #562 on qiskit-finance qiskit-community/qiskit-finance#302 but that is not draft. Maybe with the new theme things look better. Although personally having the lists/indices of attributes and functions  with the one liner makes it easy to browse things quickly to find something you might be looking for. Now I would have to to scroll the entire page (and to me the functions do not stand out so well given they and the parameters are all highlighted similarly) - or glean enough info from the name in the right sidebar to perhaps go look at that. And speaking of right sidebar - why is the width of the page not take up fully e.g. in this screenshot

image
 
I see its the same with Qiskit - though at least it shows a scollbar there probably as it using the new theme whereas the docs build from this CI do not yet. Even so though it seems to behave the same and does not use extra space that is there which could allow the wrapped names to be on one line instead of having to wrap them which is made worse by all being prefixed by the class name.

Eric-Arellano:
Hey @woodsp-ibm

> Now https://github.com/qiskit-community/qiskit-optimization/pull/562 is not yet merged - was the intent to do this first.

This PR is pre-work for #562. Without this PR, switching to the new Ecosystem Theme is prohibitively slow because Furo has O(n^2) time complexity for docs build time, with n being number of pages. This was explored in https://github.com/Qiskit/qiskit_sphinx_theme/issues/328, including with the Furo maintainer.

The intent is to merge both this PR and #562. Arnau was only splitting this up into a dedicated PR so that code review is simpler. If your team prefers, they could be combined into one PR.

For other projects, n is small enough in the O(n^2) that switching to the new theme does not require this API docs reorganization. That's why some PRs aren't in draft mode.

> Maybe with the new theme things would look better here. 

Yeah, we had the same experience when switching Qiskit itself. The new theme makes the API pages much better, e.g. https://qiskit.org/documentation/stubs/qiskit.circuit.QuantumCircuit.html#qiskit.circuit.QuantumCircuit. The right side bar works better, and we worked with the design team to improve the design.

> which is made worse by all being prefixed by the class name.

Unfortunately there's no fix for that to my knowledge.

woodsp-ibm:
@Eric-Arellano I am ok with the PRs as-is to save messing around, though it makes it harder to see what the final outcome is going to be since this is still using the older theme, hence the comments. But I guess we can bear with it now and we will see in  the theme change PR, after this is merged and that updated, how it all looks. And if anything is needed further sort it then

As to the right hand sidebar. Things do not extend across the page in the API ref - at least what is published with Qiskit though perhaps that's not off the very latest theme either. E.g. this from QuantumCircuit below where you can see the menu bar above extends to the width of the browser but the API ref. contents do not, so the right hand bar, and you can see the scrollbar for it, has a bunch of white space to the right that if used would help allieviate the method name wrapping.

image

I think this sidebar will become more important to getting an overview of the class as the table, which provided that before and arguably better usability in that regard, is not longer created. I pretty much ignored that sidebar before but it seems like the only option now oing forwards if I want to quickly scope methods to narrow in on something that might do for the task I have in mind - aside from scrolling the entire page content. The sidebar is not as good imho in regard of providing such an overview as the prior table. since the latter had the one liner description in there too. But I guess we can see how it goes...

Eric-Arellano:
> I am ok with the PRs as-is to save messing around, though it makes it harder to see what the final outcome is going to be since this is still using the older theme, hence the comments.

Sg. I'm sure Arnau would be glad to update the Ecosystem Theme PR tomorrow to be built on top of this one so that you can get the final preview :) 

>  Things do not extend across the page in the API ref 

Yeah, that page layout was inherited by Furo. I spent some time this summer investigating changing the right side bar to always float to the right, but it was non-trivial so we decided to not prioritize it. Definitely doable, only a question of time. Tracked by https://github.com/Qiskit/qiskit_sphinx_theme/issues/394. To set clear expectations, it's unlikely I'd have time to pick up that improvement since focus is on the new docs at https://docs.quantum-computing.ibm.com

woodsp-ibm:
>  ...changing the right side bar to always float to the right,

I was more thinking of it being variable width, beyond some min value, so it took up that extra space and hopefully helped avoid some of the wrapping of the names it contained.

woodsp-ibm (approved):
Approving based on above discussion.

Eric-Arellano:
> I was more thinking of it being variable width, beyond some min value, so it took up that extra space and hopefully helped avoid some of the wrapping of the names it contained.

Ah, Furo's design is for it to always be the same width, even on mobile view where it uses a slideout menu.

> Approving based on above discussion.

Thanks, Steve! Arnau doesn't have write permissions so this will need a merge from a maintainer.

PR #569, opened 2023-11-16
Improve documentation of INFINITY

### LINKED ISSUE #10: Documentation of qiskit.optimization.INFINITY looks strange

`INFINITY` is a constant to represent the infinite number.
The current description seems to be strange.
> Convert a string or number to a floating point number, if possible.

https://qiskit.org/documentation/stubs/qiskit.optimization.INFINITY.html#infinity

woodsp-ibm:
### Summary

Fixes #10 

It now comes out as

image

with the linked page

image

It did try with autodata which brings what is in the separate page inline there on the main page. But then it has that yellow stripe across the page which changed things more. I know we have brought all methods onto the same page so maybe that is reasonable to do. But since it changed things more than simply fixing the text of what was there I left it with this more minimal change.

Either as it stands or with autodata it needs the fully qualified path to bring in the comment for the constant - otherwise you still get that `Convert a string or number to a floating point number, if possible.` which is what the docs currently show,

### Details and comments

I labelled it backport since we publish docs off stable and maybe we want to update/re-publish the docs with this fix.

### DIFF
--- a/qiskit_optimization/__init__.py
+++ b/qiskit_optimization/__init__.py
@@ -68,9 +68,7 @@ are that it cannot proceed to completion.
    :toctree: ../stubs/
    :nosignatures:
 
-    INFINITY
-
-A constant for infinity.
+    ~qiskit_optimization.infinity.INFINITY
 
 Submodules
 ==========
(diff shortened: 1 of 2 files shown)


t-imamichi (approved):

t-imamichi:
Thank you for fixing the longstanding issue!

PR #580, opened 2023-12-11
Add Python 3.12 support

### LINKED ISSUE #577: cannot import "BaseBackent" from "qiskit.providers"
### Environment

- **Qiskit Optimization version**: 0.6.0
- **Python version**: 3.12.0
- **Operating system**:  linux

### What is happening?

I am trying to run the quiskit optimization example from:
https://qiskit.org/ecosystem/optimization/tutorials/02_converters_for_quadratic_programs.html

And get the following import error: 

ImportError: cannot import name 'BaseBackend' from 'qiskit.providers' (/home/jdara/anaconda3/envs/quant4/lib/python3.12/site-packages/qiskit/providers/__init__.py)

### How can we reproduce the issue?

First run these pip commands: 

pip install qiskit
pip install qiskit-ibm-runtime
pip install qiskit[visualization]
pip install qiskit-optimization
pip install 'qiskit-optimization[cplex]'
pip install 'qiskit-optimization[cvx]'
pip install 'qiskit-optimization[matplotlib]'

Then run in a jupyternotebook: 

from qiskit_optimization.algorithms import CplexOptimizer
cplex_optimizer = CplexOptimizer()

This results in this error: 

ImportError: cannot import name 'BaseBackend' from 'qiskit.providers' (/home/jdara/anaconda3/envs/quant4/lib/python3.12/site-packages/qiskit/providers/__init__.py)

I am running in a anaconda environment if this should be relevant 

### What should happen?

I would expect the import not to go wrong 

### Any suggestions?

_No response_

### LINKED ISSUE #590: Add Python 3.12 support
Qiskit 0.45.1 added Python 3.12 support, and its been added to Algorithms see qiskit-community/qiskit-algorithms#108 (although at the time of writing this algorithms has yet to officially release a version supporting this).

~CI will likely need to be specific to 3.12.0 until an updated testtools is released - see qiskit-community/qiskit-algorithms#113 which goes into the failure using 3.12.1~ 3.12.2 is ok as the breaking change was reverted so the testtools update is not needed

t-imamichi:
### Summary

Closes #590

Add Python 3.12 support.
But since CPLEX does not support Python 3.12 (see [pypi](https://pypi.org/project/cplex/#files)), we need to add constraint to 'cplex' extra dependency.
Windows and mac are still tested with Python 3.11 because cplex are not available for 3.12.

Fixes #577 

### Details and comments

Latest cplex 22.1.1.1
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -116,7 +116,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.9, '3.10', 3.11]
+        python-version: [3.8, 3.9, '3.10', 3.11, 3.12]
         include:
           - os: macos-latest
             python-version: 3.8
@@ -325,6 +325,10 @@ jobs:
           path: /tmp/o311
       - uses: actions/download-artifact@v4
         with:
+          name: ubuntu-latest-3.12
+          path: /tmp/o312
+      - uses: actions/download-artifact@v4
+        with:
           name: macos-latest-3.8
           path: /tmp/m38
       - uses: actions/download-artifact@v4
@@ -344,7 +348,7 @@ jobs:
         shell: bash
       - name: Combined Deprecation Messages
         run: |
-          sort -f -u /tmp/o38/opt.dep /tmp/o39/opt.dep /tmp/o310/opt.dep /tmp/o311/opt.dep /tmp/m38/opt.dep /tmp/m311/opt.dep /tmp/w38/opt.dep /tmp/w311/opt.dep || true
+          sort -f -u /tmp/o38/opt.dep /tmp/o39/opt.dep /tmp/o310/opt.dep /tmp/o311/opt.dep /tmp/o312/opt.dep /tmp/m38/opt.dep /tmp/m311/opt.dep /tmp/w38/opt.dep /tmp/w311/opt.dep || true
         shell: bash
       - name: Coverage combine
         run: coverage3 combine /tmp/o38/opt.dat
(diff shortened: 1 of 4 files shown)


dthuerck:
Works as intended - on `3.12` , `cplex` is not installed. On `3.10`, all is good. However, shouldn't we either write this down either in the README or, even better, give a warning to users of Python 3.12.
As it is now, the package happily installs (without `cplex`) but then gives the same error as in #577.

One idea would be, e.g., to guard the cplex imports with a warning if users are on `3.12` (as long as there is no updated `cplex` package).

dthuerck (approved):
See comment above - works, but user experience might be better served with adding warnings.

t-imamichi:
If cplex does not exists, we already display a warning message with `@_optionals.HAS_CPLEX`. It might be nice to elaborate the message to mention that cplex is not available for python 3.12 at this moment.
https://github.com/qiskit-community/qiskit-optimization/blob/36ad1a13984403e77a5308fc38d61ee1a1a1637d/qiskit_optimization/algorithms/cplex_optimizer.py#L28-L29

woodsp-ibm:
Arguably qiskit-optimization does not support 3.12 yet. Qiskit 0.45.1 added 3.12 support. I added support to qiskit algorithms qiskit-community/qiskit-algorithms#108 such that when this is released the apps, like Optimization here, could be updated to inlcude 3.12 as well.

t-imamichi:
Yes, it makes sense to include this PR's change as part of a separate PR to add support for Python 3.12.

t-imamichi:
@woodsp-ibm I added Python 3.12 support and CI. Could you take a look at it?

woodsp-ibm:
As we are adding 3.12 to the list of supported versions in setup can we add a reno like I did in ML for the 3.12 support, see qiskit-community/qiskit-machine-learning#774 (its the same basic test as what I did in Nature and Algorithms!)

t-imamichi:
Thanks. I added reno.

PR #581, opened 2023-12-11
Add mps support

dthuerck:

### Summary
See issue  #582 - adds support for `mps` and `mps.gz, .lp.gz` files through `docplex`.

### Details and comments
`gzip` is used to extract files on the fly. I replicated the single test that uses `lp` files and modified it so that it uses `mps` files.


dthuerck:
Moving this back to `draft` as there is still a permission error with temporary files under windows.

t-imamichi:
Since `QuadraticProgram` (QP) code is already big, we carefully add methods to it. It increases the maintenance cost.
qiskit_optimization has modules to separate functionality as follows and I suggest take advantage of it.

- qiskit_optimization.translators: translate between QP and other objects
- qiskit_optimization.converters: convert QP to QP in another form.

For this PR's purpose, I suggest making functions `qiskit_optimization.translators.{from_mps,to_mps}` in a separate file such as `mps.py` in `qiskit_optimziation/translators`.

`QP` has methods `read_from_lp_file/write_to_lp_file` before we introduce the translators. So, it might be nice to make `from_lp_file` `to_lp_file` functions to refactor and simplify `QP` codes in `quadratic_program.py`.

dthuerck:
As discussed in #581, will replace this PR with another one.

Issue #582, opened 2023-12-11
Add mps/gz support


dthuerck:
### What should we add?

What should we add?

Most collections of optimization models use the old-fashioned MPS format for model storage, e.g., [MIPLIB](https://miplib.zib.de/). Furthermore, due to their verbosity (compare, e.g., square47.mps.gz wit squaren47.mps from MIPLIB - 80 MB vs. 1.4GB) people tend to use the gzipped files directly.

If qiskit optimization serves as gateway to classical and quantum optimization, I think it makes perfect sense to include this widely-adopted file format. docplex supports mps natively.

t-imamichi:
Users can import and export docplex models via `from_docplex_mp` and `to_docplex_mp`. So, I think they can handle MPS format through docplex model.
We carefully think of adding new public methods to `QuadraticProgram` because it increases the maintenance cost.

Adding new methods to `QuadraticProgram`

Pros
- can directly handle MPS by `QuadraticProgram`

Cons
- need to handle MPS via docplex model and `from_docplex_mp` and `to_docplex_mp`.
- increase maintenance cost of `QuadraticProgram`.
- future docplex changes may affect the correctness of the new methods.

t-imamichi:
As commented in https://github.com/qiskit-community/qiskit-optimization/pull/581#issuecomment-1851326579, I personally suggest making functions in `qiskit_optimizatiopn.translators`.

dthuerck:
Sure, we can move it there @t-imamichi . In this case, I'd vote that we also move all `lp`-related functions (as you suggested in the PR).

Since this breaks the `QuadraticProgram`, we could add a deprecation warning and in the meantime, forward to the functions in `translator`. What do you think?

woodsp-ibm:
> we could add a deprecation warning and in the meantime, forward to the functions in translator. What do you think?

For breaking API changes maintaining compatibility via deprecation, which informs the replacement (or whatever) to maintain users code working and give them a chance to change, is definitely the way to go. Works for me.

The deprecation would also be noted in a [release note](https://github.com/qiskit-community/qiskit-optimization/blob/main/CONTRIBUTING.md#release-notes) `deprecations` section as part of the PR so it becomes part of the compete release notes in end-user docs. Since the PR is also adding new features these should be listed there in the `features` section of the same note (other unused sections would be simply deleted from the note as creation creates one with all sections present).

dthuerck:
[...]

PR #583, opened 2023-12-27
Add MPS and GZ support in `translators`

dthuerck:

### Summary

See issue #582: This PR succeeds #581, having moved lp/mps/gz IO functions into `translators`.

### Details and comments

`lp` functions in `QuadraticProgram` have been marked as deprecated. Tests and Tutorials have been updated to use the `file_io`.


docs/tutorials/01_quadratic_program.ipynb
```diff
@@ -759,39 +759,31 @@
...
-      "\n",
-      "Bounds\n",
-      " 0 <= _e <= 1\n",
-      " 0 <= f <= 1\n",
-      "\n",
-      "Binaries\n",
-      " _e f\n",
-      "End\n",
-      "\n"
+     "ename": "NameError",
```
woodsp-ibm, line 766: It looks like perhaps just this cell was run - probably best to re-run the whole notebook locally for whats checked in. It will get re-run too when docs are built but it should be checked in a state that would match what CI produces.

qiskit_optimization/problems/quadratic_program.py
```diff
@@ -917,45 +916,47 @@ def export_as_lp_string(self) -> str:
         Returns:
             A string representing the quadratic program.
         """
+        warn(
```
woodsp-ibm, line 919: There are some deprecation decorators  in Qiskit that we try and use now. They also emit the deprecation into the docstring so the notices come out in the published docs. You can most likely find usage examples in Qiskit - this is the file with the decorators https://github.com/Qiskit/qiskit/blob/main/qiskit/utils/deprecation.py

woodsp-ibm (commented):
A couple of quick comments. And it will need a release note for the new function and to note the deprecation see https://github.com/qiskit-community/qiskit-optimization/blob/main/CONTRIBUTING.md#release-notes for info on this, if you need it, which is done with a tool called reno.

docs/tutorials/01_quadratic_program.ipynb
dthuerck: Thanks - re-ran and checked it.

qiskit_optimization/problems/quadratic_program.py
dthuerck: Thanks again for the reference - used the appropriate decorators in the last commit.

dthuerck:
@woodsp-ibm Thanks for the hints! Should be all done by now, release note is there.

woodsp-ibm (commented):
One other thing.  Its now 2024 and many of these files still have 2023 dates. This passes on CI presently but, from past experience, the files, when merged in to main will end up have last updated dates of 2024. This PR will not fail, but any subsequent ones, since the checks are run against all files, will fail on these despite it might not even touch them. To avoid this, if I could ask you to update the dates for 2024 in all the files with a copyright that will avoid us having to deal with it down the line. Its a bit of a pain I know but having a PR cross a year boundary has been more of an exception.

dthuerck:
Thanks for the extensive comments, @woodsp-ibm ! I added test cases for the deprecated functions and updated the docs accordingly.

Issue #591, opened 2024-02-05
Documentation contains outdated CPLEX version 


larsschellhas:
### Environment

Irrelevant

### What is happening?

The documentation states "Currently there is no python **3.9** version of CPLEX. In this case, the CPLEX install command will have no effect."

### How can we reproduce the issue?

See Readme.md

### What should happen?

It should say "Currently there is no python **3.12** version of CPLEX. In this case, the CPLEX install command will have no effect."

woodsp-ibm:
Fixed by #592 which has now been merged.

PR #592, opened 2024-02-05
Update CPLEX-incompatible Python version in documentation 

larsschellhas:
### Summary
Updates the version number of Python for which no CPLEX version exists, yet.
In this case, updating from 3.9 to 3.12.

This fixes issue #591.

### Details and comments

No more details required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,7 @@ To do this follow the instructions in the
 ### Optional Installs
 
 * **IBM CPLEX** may be installed using `pip install 'qiskit-optimization[cplex]'` to enable the reading of `LP` files and the usage of
-  the `CplexOptimizer`, wrapper for ``cplex.Cplex``. Currently there is no Python 3.12 version of CPLEX. In this case, the CPLEX install
-  command will have no effect.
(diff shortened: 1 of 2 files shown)


woodsp-ibm:
Given something specific will always be out of date maybe it needs to be a more generic statement there. As it stands Qiskit Optimization also does not officially support 3.12 yet, see #590 - though I imagine its likely it works.

So perhaps instead of something specific, i.e.

`Currently there is no Python 3.12 version of CPLEX. In this case, the CPLEX install command will have no effect.`

Perhaps it ought to be changed to something more like

`CPLEX is a separate package and its support of Python versions is independent of Qiskit Optimization, where this CPLEX command will have no effect if there is not (yet) a compatible version of CPLEX available.`

larsschellhas:
@woodsp-ibm, of course, makes sense. I've updated the respective sections.

t-imamichi (approved):
Thank you for your contribution. LGTM.

woodsp-ibm (approved):
Thanks for the contribution in improving the docs.

PR #595, opened 2024-02-05
Update actions using deprecated Node.js

woodsp-ibm:
### Summary

Currently CI - if you look in the build details Summary - is emitting a bunch of deprecation warnings around actions and Node.js

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-python@v4, actions/cache@v3, actions/upload-artifact@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

This updates the actions to the latest versions.

### Details and comments

I will mark this as stable backport - I think the workflows should backport ok as it has not been long since the last release and it would be better to have them updated as well.

Note: I have updated the actions in the other apps (Nature and ML, as well as Algorithms). I did this manually here for all the actions but intend to add a dependabot config to have it create PRs, when needed in the future for updates. It creates a PR for each action, which is the reason I went ahead and did all actions here as I found out that the upload and download needed to be done together for things to work (it errored finding the file if there was a version mismatch) so I ended up there doing a manual PR anyway to combine the upload and download hence I figured I would have to at least do that here so I may as well do the lot, add the dependabot config afterwards as a separate PR so the updating can have some automation for the future.

### DIFF
--- a/.github/actions/install-main-dependencies/action.yml
+++ b/.github/actions/install-main-dependencies/action.yml
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2021, 2023.
+# (C) Copyright IBM 2021, 2024.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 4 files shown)


t-imamichi (approved):
LGTM

PR #596, opened 2024-02-05
Add dependabot to manage future Github Actions updates

woodsp-ibm:
### Summary

This adds a `dependabot` config in order that the actions updates are automated for the future, as I have done in the apps repos (ML and Nature) as well as the Algorithms. 

### Details and comments

I labelled this `On Hold` as ideally #595 should be merged first and then this config afterwards.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# Check for updates to GitHub Actions
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


t-imamichi:
stable backport?

woodsp-ibm:
> stable backport?

@t-imamichi `dependabot` works against the default branch and will do PRs against that so I don;t think backporting it has any real value. Any PRs dependabot creates can always be labelled stable backport so they are backported by mergify when merged. Eventually this config will end up on a stable branch, when a new minor is released, but we don't need to backport it presently - we could if it was really wanted just to have it there too, buts it not needed afaik.

t-imamichi:
Thank you for the details. I merge it without stable backport.

t-imamichi (approved):
LGTM

PR #605, opened 2024-02-13
Remove qiskit.org links from release notes

Eric-Arellano:
Second part of https://github.com/qiskit-community/qiskit-optimization/pull/604


woodsp-ibm (approved):
Thx again Eric.

FYI for others this, just release note changes, was separated out from the other link schanges as there is logic in the stable branch which checks release notes last updated date and whether that is after a tagged release. If so it does not include the release note(s) in question. This allows us to backport fixes ready for a release yet at anytime publish docs updates - if we did not have something like this then the docs would included notes for as yet unreleased content.

PR #609, opened 2024-02-16
Pin qiskit<1.0

t-imamichi:
### Summary

CI fails with Qiskit 1.0. https://github.com/qiskit-community/qiskit-optimization/actions/runs/7924881956
This PR is a workaround until qiskit-algorithms 0.3.0 is released.
I locally confirmed that qiskit-optimization main + qiskit-algorithms main + qiskit 1.0 pass all tests.
More specifically, we need https://github.com/qiskit-community/qiskit-algorithms/pull/98.

### DIFF
--- a/.github/actions/install-main-dependencies/action.yml
+++ b/.github/actions/install-main-dependencies/action.yml
@@ -81,7 +81,7 @@ runs:
             source "$CONDA/etc/profile.d/conda.sh"
             conda activate scsenv
           fi
-          pip install -U qiskit
+          pip install -U 'qiskit<1'
         fi
       shell: bash
     - name: Install stable Aer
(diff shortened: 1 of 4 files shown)


woodsp-ibm:
So a question. I just fixed up tutorials for the algorithms as CI was failing there qiskit-community/qiskit-algorithms#156. Yesterday in Finance a backport failed, it was done just after 1.0 released where the original PR was done with 0.46, and I just fixed them the same way - CI is now failing there the same way as here, now that the tutorials get further. I need to fix the tutorials here too and that needs backporting - and that would fail unless this is backported too or we just leave that failing on stable for now. The question is do you just want to wait until algorithms is released so this does not have to be merged on main and then reverted in a couple of days.

woodsp-ibm:
@t-imamichi Qiskit Algorithms 0.3.0 has been released so this pinning is no longer needed. The PR fixing the tutorials #610 is now passing, except for test of the stable tutorials which 610 is needed to be backported to fix.

woodsp-ibm:
I am going to close this off since its no longer needed, now that Qiskit Algorithms 0.3.0 has been released.

PR #610, opened 2024-02-16
Fix tutorials to work with Qiskit 1.0

woodsp-ibm:
### Summary

Qiskit 1,0 removed the deprecated qiskit.tools which included some .jupyter from which the magics that did the version and copyright were imported. This PR adds a local file in the tutorials folder that contains just these magics and changes the import to use this.

Similar fix to qiskit-community/qiskit-algorithms#156

I removed an import from qiskit.tools which is now gone - the import did not seem to be used anymore

I also tweaked one of the imports so it does not import direct from the module but just from qiskit_algorithms.utils

### Details and comments

I do not expect this to pass until the other failures are sorted. A new release of Qiskit Algorithms is required to work with Qiskit 1.0, but at least its ready.

### DIFF
--- a/docs/tutorials/01_quadratic_program.ipynb
+++ b/docs/tutorials/01_quadratic_program.ipynb
@@ -825,7 +825,7 @@
     }
    ],
    "source": [
-    "import qiskit.tools.jupyter\n",
+    "import tutorial_magics\n",
     "\n",
     "%qiskit_version_table\n",
     "%qiskit_copyright"
--- a/docs/tutorials/02_converters_for_quadratic_programs.ipynb
+++ b/docs/tutorials/02_converters_for_quadratic_programs.ipynb
@@ -697,7 +697,7 @@
     }
    ],
    "source": [
-    "import qiskit.tools.jupyter\n",
+    "import tutorial_magics\n",
     "\n",
     "%qiskit_version_table\n",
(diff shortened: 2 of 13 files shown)


woodsp-ibm:
@t-imamichi This is ready to go - we will have to bypass the branch protections for the failure of the stable tutorials (they will not work until this is backported as the copyright/version does not exist in qiskit.tools anymore).

t-imamichi:
Tutorial test with Python 3.8 fails. Could you take a look at it?

woodsp-ibm:
> Tutorial test with Python 3.8 fails. Could you take a look at it?

@t-imamichi As I commented above its the testing of the tutorials that are in the stable branch. We do this just on 3.8 and its done to make sure they still work with whatever changes are done in main to ensure compatibility. But in this case its not a change in main but in Qiskit and they are going to fail since they still have the former copyright and version that this PR changes since qiskit.tools no longer exists in Qiskit 1.0. This PR will need to be merged as-is which will require bypassing the branch rule as I mentioned - once backported to stable it will fix them there. (This was already done for Nature and ML where it was the same thing - you can see in Finance, that also has yet to be done, its the same too)

t-imamichi (approved):
Sorry. I missed your comments above. LGTM.

PR #627, opened 2024-08-01
Fix incorrect decoding of decision variables by ``MagicRounding``

### LINKED ISSUE #626: Solving a bug in the MagicRounding class of QRAO
### Environment

- **Qiskit Optimization version**: 0.6.1
- **Python version**: 3.10
- **Operating system**: Ubuntu 20.04

### What is happening?

The problem arises in the method, ``_unpack_measurement_outcome,'' in MagicRounding that decodes the classical bits from a quantum state.

In the current version, we can find in lines 234-235:
``` 
#iterate in order over decision variables
for q, op in var2op.values():
```
This code iterates in the order in which values were added to the dictionary, rather than in the order of the decision variables.
So, the line 235 should be modified as follows:
```diff
# iterate in order over decision variables
- for q, op in var2op.values():
+ for var in range(len(var2op)):
+     q, op = var2op[var]
```

### How can we reproduce the issue?

In the following, the maxcut problem for the one-dimensional graph with 3 nodes is solved.
The maxcut value should be 2, but the result returns the cut value of 1.

```
import networkx as nx
from qiskit import QuantumCircuit
from qiskit.primitives import Sampler
from qiskit_optimization.applications import Maxcut
from qiskit_optimization.algorithms.qrao import (
    QuantumRandomAccessEncoding,
    MagicRounding,
)
from qiskit_optimization.algorithms.qrao.rounding_common import RoundingContext


# Create one-dimensional graph whose maxcut is trivial
graph = nx.Graph()
graph.add_nodes_from([0, 1, 2])
graph.add_edges_from([(0, 1), (1, 2)])
maxcut = Maxcut(graph)
problem = maxcut.to_quadratic_program()
print(problem.prettyprint())


# Encode the QUBO problem into a Hamiltonian
encoding = QuantumRandomAccessEncoding(max_vars_per_qubit=1)
encoding.encode(problem)


# Construct a quantum circuit that prepares the optimal state.
circuit = QuantumCircuit(3)
circuit.x(2-encoding.var2op[1][0]) # since the endianness of var2op and the endianness of the qubit numbers are reversed.


# Perform magic state rounding
rounding_context = RoundingContext(encoding=encoding, expectation_values=0, circuit=circuit)
sampler = Sampler(options={"shots": 1})
magic_rounding = MagicRounding(sampler=sampler)
results = magic_rounding.round(rounding_context)
print(results.samples[0])
```

### What should happen?

The above code should returns the maxcut value of 2.

### Any suggestions?

Modifying the lines 234-235 of MagicRounding will fix the bug.
It is also preferable to align the endianness of var2op with the endianness of the qubit numbers.

yksat:
### Summary
Fix ``MagicRounding`` so that the decision variables can be decoded correctly.
Fix #626. 

### Details and comments
The iteration order of ``var2op`` in the method, ``_unpack_measurement_outcome`` in :class:`~qiskit_optimization.algorithms.qrao.MagicRounding` has been modified so that it can correctly run in the order of the decision variables.

### DIFF
--- a/releasenotes/notes/fix-qrao-d0402da6a5c40121.yaml
+++ b/releasenotes/notes/fix-qrao-d0402da6a5c40121.yaml
@@ -1,5 +1,4 @@
 ---
 fixes:
   - |
-    Fixed incorrect iteration order ``var2op.values()``
-    in :class:`~qiskit_optimization.algorithms.qrao.MagicRounding`.
+    Fixed a bug of :meth:`.MagicRounding.round` that may return a wrong decision variable order.
--- a/releasenotes/notes/fix-qrao-d0402da6a5c40121.yaml
+++ b/releasenotes/notes/fix-qrao-d0402da6a5c40121.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixed incorrect iteration order ``var2op.values()``
+    in :class:`~qiskit_optimization.algorithms.qrao.MagicRounding`.
--- a/qiskit_optimization/algorithms/qrao/magic_rounding.py
+++ b/qiskit_optimization/algorithms/qrao/magic_rounding.py
@@ -232,7 +232,8 @@ class MagicRounding(RoundingScheme):
         """
         output_bits = []
         # iterate in order over decision variables
-        for q, op in var2op.values():
+        for var in sorted(var2op.keys()):
+            q, op = var2op[var]
             # get the decoding outcome index for the variable
             # corresponding to this Pauli op.
             op_index = self._OP_INDICES[vars_per_qubit][str(op.paulis[0])]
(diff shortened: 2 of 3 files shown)


releasenotes/notes/fix-qrao-d0402da6a5c40121.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixed incorrect iteration order ``var2op.values()``
+    in :class:`~qiskit_optimization.algorithms.qrao.MagicRounding`.
```
t-imamichi, line 5:
```suggestion
    Fixed a bug of :meth:`.MagicRounding.round` that may returns a wrong solution order.
```

t-imamichi (approved):
Thank you for your contribution. LGTM

PR #630, opened 2024-08-05
Pin pip for "pip check"

t-imamichi:
### Summary

`pip check` of CI fails since a couple of days ago.
```
gurobipy 11.0.3 is not supported on this platform
```

I found an issue of pip https://github.com/pypa/pip/issues/12884 is likely to be related to the failure. I try to pin pip only for `pip check`.


woodsp-ibm (approved):
LGTM - thx for linking the pip issue, was helpful to understand

Issue #611, opened 2024-02-19
Solving a bug in the MaxCut class


Maldini32:
### Environment

- **Qiskit Optimization version**: 0.45.1
- **Python version**: 3.10
- **Operating system**: Windows 10

### What is happening?

The problem regards the method parse_gset_format, which should be updated in order to correctly read the files.

While in the library version we can find this content in lines 117 to 120:

```
s__, t__, _ = v
s__ -= 1  # adjust 1-index
t__ -= 1  # ditto
w[s__, t__] = t__
```

They should be replace with these lines in order to properly assign the weight to the edges:

```
s__, t__, _ = v
s__ -= 1  # adjust 1-index
t__ -= 1  # ditto
w[s__, t__] = _
```

### How can we reproduce the issue?

The problem is solved by modifying the 120 to:

```
w[s__, t__] = _
```

### What should happen?

By modifying this line the weights will be correctly assigned to the problem instance.

MiasWuQG:
Thank you for identifying this issue and providing a detailed explanation of the required changes. To help resolve this more efficiently, please consider submitting a Pull Request (PR) with your suggested code modifications. This will allow us to review, test, and merge the fix quickly. Let us know if you need any assistance with the PR process!

PR #635, opened 2024-10-08
Fix bug in max-cut solver (issue #611)

TolisChal:
This PR fixes the bug in the max-cut solver as reported in https://github.com/qiskit-community/qiskit-optimization/issues/611

### DIFF
--- a/qiskit_optimization/algorithms/admm_optimizer.py
+++ b/qiskit_optimization/algorithms/admm_optimizer.py
@@ -42,7 +42,7 @@ logger = logging.getLogger(__name__)
 class ADMMParameters:
     """Defines a set of parameters for ADMM optimizer."""
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-positional-arguments
         self,
         rho_initial: float = 10000,
         factor_c: float = 100000,
@@ -175,7 +175,7 @@ class ADMMState:
 class ADMMOptimizationResult(OptimizationResult):
     """ADMMOptimization Result."""
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-positional-arguments
         self,
         x: np.ndarray,
         fval: float,
--- a/qiskit_optimization/algorithms/admm_optimizer.py
+++ b/qiskit_optimization/algorithms/admm_optimizer.py
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2020, 2023.
+# (C) Copyright IBM 2020, 2024.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
--- a/qiskit_optimization/applications/max_cut.py
+++ b/qiskit_optimization/applications/max_cut.py
@@ -114,10 +114,10 @@ class Maxcut(GraphOptimizationApplication):
                     w = np.zeros((n, n))
                     header = False
                 else:
-                    s__, t__, _ = v
+                    s__, t__, w__ = v
                     s__ -= 1  # adjust 1-index
                     t__ -= 1  # ditto
-                    w[s__, t__] = _
+                    w[s__, t__] = w__
                     count += 1
             assert m == count
         w += w.T
--- a/qiskit_optimization/applications/max_cut.py
+++ b/qiskit_optimization/applications/max_cut.py
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2018, 2023.
+# (C) Copyright IBM 2018, 2024.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
--- a/qiskit_optimization/applications/max_cut.py
+++ b/qiskit_optimization/applications/max_cut.py
@@ -117,7 +117,7 @@ class Maxcut(GraphOptimizationApplication):
                     s__, t__, _ = v
                     s__ -= 1  # adjust 1-index
                     t__ -= 1  # ditto
-                    w[s__, t__] = t__
+                    w[s__, t__] = _
                     count += 1
             assert m == count
         w += w.T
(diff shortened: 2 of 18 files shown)


t-imamichi:
Thank you for fixing an issue.
I read the changelog of pylint and found that a new check `too-many-positional-arguments` was introduced by pylint 3.3.
https://pylint.readthedocs.io/en/latest/whatsnew/3/3.3/index.html#new-checks

Because qiskit-optimization already has the following config for `max-args`, how about adding a similar config for `max-positional-args` as follows too?
https://github.com/qiskit-community/qiskit-optimization/blob/807d48167caf11cd93bec85f26b34614fb7868da/.pylintrc#L313

```
max-positional-arguments=8
```
https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/too-many-positional-arguments.html

t-imamichi:
Let me try the config #637

t-imamichi:
How about merging #637? You can minimize the change with the config.

t-imamichi:
I closed #637 following @woodsp-ibm' comment.

qiskit_optimization/algorithms/admm_optimizer.py
```diff
@@ -42,7 +42,9 @@
 class ADMMParameters:
     """Defines a set of parameters for ADMM optimizer."""
 
-    def __init__(
+    # Delete all pylint checks because (too-many-positional-arguments) is unknown
+    # option value in some platforms in github CI
+    def __init__(  # pylint: disable=all
```
t-imamichi, line 47: `disable=all` is too broad.
How about the following?
```suggestion
    def __init__(  # pylint: disable=too-many-positional-arguments
```

qiskit_optimization/applications/max_cut.py
```diff
@@ -117,7 +117,7 @@ def parse_gset_format(filename: str) -> np.ndarray:
                     s__, t__, _ = v
                     s__ -= 1  # adjust 1-index
                     t__ -= 1  # ditto
-                    w[s__, t__] = t__
```
t-imamichi, line 120: Since `_` is usually used for non-used value, I suggest renaming it, e.g., `w__`.
```suggestion
                    s__, t__, w__ = v
                    s__ -= 1  # adjust 1-index
                    t__ -= 1  # ditto
                    w[s__, t__] = w__
```

t-imamichi:
LGTM overall. I wrote some comments.

qiskit_optimization/algorithms/admm_optimizer.py
```diff
@@ -175,7 +177,9 @@ def __init__(self, op: QuadraticProgram, rho_initial: float) -> None:
 class ADMMOptimizationResult(OptimizationResult):
     """ADMMOptimization Result."""
 
-    def __init__(
+    # Delete all pylint checks because (too-many-positional-arguments) is unknown
+    # option value in some platforms in github CI
+    def __init__(  # pylint: disable=all
```
t-imamichi, line 182: same as https://github.com/qiskit-community/qiskit-optimization/pull/635/files#r1802386822

qiskit_optimization/algorithms/admm_optimizer.py
TolisChal: If I use `# pylint: disable=too-many-positional-arguments` in some platform tests I'm getting an error that `too-many-positional-arguments` is an unknown pylint flag. I guess this happens because some of the platforms doesn't have the updated version of the pylint.

qiskit_optimization/algorithms/admm_optimizer.py
TolisChal: Please read my answer above.

qiskit_optimization/applications/max_cut.py
TolisChal: Thanks! I changed it as you proposed.

qiskit_optimization/algorithms/admm_optimizer.py
woodsp-ibm: That happens in Python 3.8 as the newer pylint is not supported there and an older version of pylint that is compatible gets installed that does not have that new option. What we did in other applications and algorithms was to add  "unknown-option-value" to pylint config to disable the error you get in that case from this new option. For instance see this config https://github.com/qiskit-community/qiskit-algorithms/blob/8709f009b61b1c573d802b4b3f8201136a29156a/pyproject.toml#L55 though elsewhere we unified much of the config in a pyproject.toml rather than having the separate config files but I would hope it should work the same way. For instance this is the PR in Machine learning that did the equivalent (seems the change to have the config in pyproject was done in that same PR)

thkleinert (dismissed):

qiskit_optimization/algorithms/admm_optimizer.py
TolisChal: Thanks @woodsp-ibm! I added both `unknown-option-value` and `too-many-positional-arguments` in the `.pylintrc` file.

t-imamichi: I guess Steve suggested adding only `unknown-option-value` to pylintrc and changing `pylint: disable` for each `__init__`. Could you confirm it, @woodsp-ibm? 

His link adds only `unknown-option-value` https://github.com/qiskit-community/qiskit-algorithms/blob/8709f009b61b1c573d802b4b3f8201136a29156a/pyproject.toml#L55
qiskit-algorithms has `pylint: disable=too-many-positional-arguments` for each file.
https://github.com/qiskit-community/qiskit-algorithms/blob/8709f009b61b1c573d802b4b3f8201136a29156a/qiskit_algorithms/phase_estimators/ipe.py#L62

woodsp-ibm: Yes, elsewhere we added the disable too many args on each failing line rather than disabling it globally with a view that for any new code we ought to do better so that this does not get raised. And as you say you then need the unknown option value disable to suppress the error you then get from the older pylint used in 3.8 .

TolisChal: Thanks guys! I modified the code accordingly! I hope I did it correctly this time :P

qiskit_optimization/applications/max_cut.py
t-imamichi: My suggestion does not seem to be applied.
Line 120 is stil
```
                    w[s__, t__] = _
```

t-imamichi (approved):
LGTM

PR #637, opened 2024-10-15
Add `max-positional-args` config

t-imamichi:
### Summary

Pylint 3.3 introduced a new check `too-many-positional-arguments`.
https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/too-many-positional-arguments.html

This PR adds a config for the check.


woodsp-ibm:
On other apps we disabled the too many positional on the line it occurred. Any new code that was done with so many positional args, would get warned against so it could be improved - while setting the config this way it would just be accepted. I know in newer code we had tried to do better with some pos args and then using * to enforce keyword only. But yes it does have an issue with the older lint that is only supported in 3.8. For that one needs to add "unknown-option-value" to the list of lint errors in teh config that is ignored. Soon qiskit will drop support for 3.8 anyway and that then would be unnecessary.

t-imamichi:
Thank you for your feedback, @woodsp-ibm. I then close this PR and review #635 so that we merge it soon.

PR #639, opened 2024-10-17
Drop python 3.8 support, add python 3.12 tests, and pin qiskit<2

t-imamichi:
### Summary

Addresses #638 
Fixes #657

qiskit-machine-learning also dropped the support of Python 3.8 already.
https://github.com/qiskit-community/qiskit-machine-learning/pull/826

This adds tests of python 3.12 for Linux and Windows because the latest CPLEX provides wheels for the platforms.
Note that CPLEX does not provide a wheel for Python 3.12 on macOS.
  - https://pypi.org/project/cplex/22.1.2.0/#files (only Windows and Linux)

This also simplifies gurobipy constraints because gurobipy provides arm64 wheels.
  - https://pypi.org/project/gurobipy/12.0.1/#files

PR #654, opened 2025-02-13
[WIP] Remove qiskit-algorithms dependency

t-imamichi:
### Summary

WIP. This PR removes qiskit-algorithms dependency based on #639.
After #639 is merged, this PR needs to be rebased onto main branch.


qiskit_optimization/optimizers/__init__.py
```diff
@@ -11,23 +11,20 @@
...
-================================================
+Optimizers (:mod:`qiskit_optimization.optimizers`)
+==================================================
 Classical Optimizers.
 
 This package contains a variety of classical optimizers and were designed for use by
-qiskit_algorithm's quantum variational algorithms, such as :class:`~qiskit_algorithms.VQE`.
+qiskit_optimization's quantum variational algorithms, such as
+:class:`~qiskit_optimization.minimum_eigensolvers.SamplingVQE`.
 Logically, these optimizers can be divided into two categories:
```
woodsp-ibm, line 21: > Logically, these optimizers can be divided into two categories:

Given that one of the two categories i.e. Global Optimizers has been removed by the changes here the (above) text really needs redoing.

woodsp-ibm:
In looking at #662 and doing a quick search on this repo I see that tutorial12 ie 12_quantum_random_access_optimizer.ipynb in line[13] when viewing it here has the following which referred to Qiskit Algorithms - but the url is not where Algorithms docs are published. But given the class has been moved (copied) here the reference needs to be local now.

MinimumEigensolverResult: This object contains the results of running the minimum eigen optimizer such as the VQE on the relaxed problem. It provides information about the eigenvalue, and other relevant details. You can refer to the Qiskit Algorithms [documentation](https://docs.quantum.ibm.com/api/qiskit/qiskit.algorithms.MinimumEigensolverResult) for a comprehensive explanation of the entries within this object.

t-imamichi:
The tutorial test currently fails on the stable branch due to its dependency on qiskit-algorithms. Removing this dependency constitutes a breaking change.
Supporting multiple versions of classes like SamplingVQE - which exist in both qiskit-algorithms and qiskit-optimization (this PR) - adds significant complexity. Given this, I propose we allow the breaking change and proceed with removing qiskit-algorithms.
If this approach is acceptable, I can move forward with updating the codebase to support Qiskit 2.0.

woodsp-ibm (approved):
Looks fine to me

t-imamichi:
Thank you!

PR #660, opened 2025-04-21
Add Python 3.13 support

### LINKED ISSUE #659: Python 3.13 support
### What should we add?

Add support for Python 3.13.

t-imamichi:
### Summary

Fixes #659 
Need to update requirements.txt to support Python 3.13 because SciPy 1.13 is not available for Python 3.13.

If this PR is approved, I will make ubuntu-latest 3.13 required for CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,7 +119,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.9, '3.10', 3.11, 3.12]
+        python-version: [3.9, '3.10', 3.11, 3.12, 3.13]
         include:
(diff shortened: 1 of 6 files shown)


woodsp-ibm (approved):
LGTM. Is the reason for leaving the Mac and Win upper versions as they were as that perhaps relates to those are the latest Python versions for which CPlex package is supported? Maybe same for the upper version of tutorials as I do not see a 3.13 version out there for that at all, so I guess the unbuntu 3.13 is done without that - which is fine I guess since ubuntu covers all the python versions and cplex is supported on the rest.

t-imamichi:
Exactly. Mac, Win, Tutorial python versions are kept because CPLEX is not available for python 3.13.

PR #662, opened 2025-05-21
Update intersphinx mapping and other URLs pointing to IQP Classic

arnaucasau:
The documentation source of truth is moving to https://quantum.cloud.ibm.com/.


woodsp-ibm (approved):
LGTM. I did see another file which has a docs.quantum url - but that needs changing rather more given that class has been "moved" here and so I noted that in #654 that it can be later updated.

t-imamichi (approved):
LGTM

Issue #638, opened 2024-10-15
Need to update Python versions for CI


t-imamichi:
### What should we add?

Since Python 3.13 was released on Oct 7 and Python 3.8 becomes EOL, we need to update CI.
- https://www.python.org/downloads/release/python-3130/
- https://discuss.python.org/t/python-3-8-is-now-officially-eol/66983

We need to make sure the CPLEX availability. The latest release 22.1.1.2 is available for up to python 3.11.

PR #671, opened 2025-08-06
Deprecate lp file

t-imamichi:
### Summary

Deprecates `QuadraticProgram.{export_as_lp_string,read_from_lp_file,write_to_lp_file}`.
They depend on cplex package and it is the blocker to use the latest Python version for CI.
Users can still load and dump models in LP format using `{from,to}_{docplex_mp,gurobipy}`.

Once these methods are removed, `QuadraticProgram` class will no longer depend on docplex or cplex packages.

Address #638


woodsp-ibm (approved):

t-imamichi:
Thank you for reviewing it.

PR #676, opened 2025-08-17
Modernize type hints

t-imamichi:
### Summary

Modernize type hints with support of ruff.
`typing` module is not necessary in most cases since Python 3.9.
https://typing.python.org/en/latest/guides/modernizing.html

### Details and comments

I used the following config of ruff

- https://docs.astral.sh/ruff/rules/#pyflakes-f
- https://docs.astral.sh/ruff/rules/#pyupgrade-up
- https://docs.astral.sh/ruff/rules/#flake8-future-annotations-fa

```toml
line-length = 100
target-version = "py39"

select = [
    "F",
    "UP006",
    "UP007",
    "UP010",
    "UP035",
    "UP036",
    "UP037",
    "UP045",
    "FA100",
]
```

### DIFF
--- a/qiskit_optimization/algorithms/admm_optimizer.py
+++ b/qiskit_optimization/algorithms/admm_optimizer.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 import copy
 import logging
 import time
-from typing import List, Optional, Tuple, cast
+from typing import cast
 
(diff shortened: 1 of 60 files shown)


woodsp-ibm (approved):
Seems fine to me

